PR #89, opened 2012-07-06
Bypass glDetachShader and glDeleteShader when dumping state

gregory38:
Follow issue 78:

Due to a driver bug, I recreate glCreateShaderProgramv in my application but I keep the part that delete/detach the shader part. Unfortunately apitrace isn't able anymore to retrieve shader code source

I think, it would be better to bypass glDetachShader and glDeleteShader too in dumping state.

### DIFF
--- a/retrace/glretrace.py
+++ b/retrace/glretrace.py
@@ -318,7 +318,7 @@ class GlRetracer(Retracer):
             print r'    } else {'
             Retracer.invokeFunction(self, function)
             print r'    }'
-        elif function.name in ('glDeleteShader', 'glDetachShader'): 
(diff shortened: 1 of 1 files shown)


jrfonseca:
Sorry for not taking action all this time.  I was reluctant between both solutions: the complexity of a shader cache, vs the shader leak/renumbering.

But I have good news: it's not necessary to skip glDeleteShader. Skipping glDetachShader suffices.

And by not skipping glDeleteShader, my concern of shaders being massively renumbered no longer applies.  Shaders are still deleted, but merely their desctruction is deferred until the program gets deleted.

So I'll take you'r original patch. And modify it to only skip glDetachShader.

gregory38:
Cool. Thank you.

PR #320, opened 2015-02-05
glhelpers: fix glprofile.cpp compilation on VS2012

Dragnalith:
std::string was forward declarated only, so I have
include <string>

Without the fix, you got template compilation error at least for x64 plateform.


jrfonseca:
Pushed thanks.

PR #321, opened 2015-02-12
specs: Ensure IDirectDrawSurface2 and IDirectDrawSurface3 are wrapped

bgK:
Without this commit, surface->QueryInterface(IID_IDirectDrawSurface2) and surface->QueryInterface(IID_IDirectDrawSurface3) return an unwrapped interface.


jrfonseca:
Thanks. Adding IDirectDrawSurface2/IDirectDrawSurface3 makes sense.

But the changes to the GetDDInterface methods look incorrect to me -- otherwise the pointer to the DD object will not be properly wrapped/unwrapped.

jrfonseca:
I see the problem.  I've pushed your fix after tweaking the GetDDInterface prototypes.

It should still work. Let me know otherwise.

bgK:
Good catch with the GetDDInterface signature. Looks like I copy/pasted the incorrect solution.

PR #324, opened 2015-02-22
Always show microseconds in the profile heat map

dangelog:
Don't let the automatic rounding trigger here. Otherwise, inspecting
a call starting at time T >= 1 seconds will always hide microseconds
(and round to milliseconds), etc.


jrfonseca:
Pushed. Thanks.

PR #326, opened 2015-03-04
os: fix getConfigDir on Android

rogerzanoni:
getenv("HOME") always return null on android and the homeDir variable
assertion will always abort the program. This change makes it look for
the home directory in the users password database if the directory is
not in the HOME environment variable.

Also, don't append the ".config" directory for android, the settings are
usually stored in a folder under the /data directory.


jrfonseca:
Pushed. Thanks.

PR #342, opened 2015-05-25
Changes related to NV_path_rendering extension used by Skia

kimmok:
Hello, would these be ok for merging?


jrfonseca:
Per https://ci.appveyor.com/project/jrfonseca/apitrace/build/43/job/geayu5bb3aymoxhg MSVC is failing with

    C:\projects\apitrace\helpers\glsize.hpp(1307): error C2065: 'GL_PATH_STROKE_BOUND_NV' : undeclared identifier

I tried updating the glxext.h khronos headers but it doesn't pick it up. It looks this is some sort of old/deprecated enum? It's still on https://www.opengl.org/registry/specs/NV/path_rendering.txt though. I guess all bets are on what is or not part of `NV_path_rendering` these days.

We'll need to add this hunk:

    diff --git a/dispatch/glimports.hpp b/dispatch/glimports.hpp
    index d535c79..12cb171 100644
    --- a/dispatch/glimports.hpp
    +++ b/dispatch/glimports.hpp
    @@ -76,6 +76,7 @@
     #define GL_DEPTH_PASS_INSTRUMENT_COUNTERS_SGIX 0x8311
     #define GL_DEPTH_PASS_INSTRUMENT_MAX_SGIX 0x8312
     #define GL_PATH_SAMPLE_QUALITY_NV         0x9085
    +#define GL_PATH_STROKE_BOUND_NV           0x9086
     #define GL_PATH_STROKE_OVERSAMPLE_COUNT_NV 0x9087
     #define GL_MULTI_HULLS_NV                 0x908C
     #define GL_PROJECTIVE_2D_NV               0x9093

kimmok:
Thanks for the review. I updated the mistakes.
I am trying to get the spec issues fixed, but it of course takes a while..

jrfonseca:
Thanks for the update. Looks good afaict. I'll merge shortly.

> I am trying to get the spec issues fixed, but it of course takes a while.

I'm perfectly fine working around any short-coming of Khronos headers, either in `dispatch/glimports.hpp`, or as patches in `thirdparty\khronos`.  The problem is that, between Khronos and https://www.opengl.org/registry/specs/NV/path_rendering.txt I don't know exactly which, if any, is right.

For my own reference, could you point to me to what should be the latest version of the spec? Is it https://www.opengl.org/registry/specs/NV/path_rendering.txt ? Or is there another newer spec somewhere in nvidia's website?

kimmok:
> For my own reference, could you point to me to what should be the latest version of the spec? Is it  https://www.opengl.org/registry/specs/NV/path_rendering.txt ? 

Yeah, the spec in www.opengl.org is the most up-to-date.
    Last Modified Date:  September 9, 2014
    Version:             35

The spec is more correct than the headers, but there's some mild inaccuracies with the spec text. The token names, their values and function signatures are correct in the spec.

jrfonseca:
Pushed. Thanks.

PR #369, opened 2015-08-07
[WIP] Add a DRM render node implementation for glws

### LINKED ISSUE #368: Implement glretrace --headless using a render node on Linux
This would allow retracing without a X11 server, for example on Wayland or on a headless server.

linkmauve:
This allows eglretrace to run headless on X-less systems, like Wayland- or KMS-based ones, fixes #368.

I put the WIP tag because this PR creates a new eglretrace_gbm and I’m not sure how you would prefer to handle it having multiple backends.

I also had to disable the X11 part of getDrawableBounds() in order to compile without linking against the Xlib, another solution would be to still link against it even in the GBM code, but I’m not a big fan of that idea. A third solution would be to stub the few Xlib functions used there, to make them fail immediately.

Thank you in advance for your pointers. :)

### DIFF
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -158,6 +158,28 @@ if (ENABLE_EGL AND X11_FOUND AND NOT WIN32 AND NOT APPLE AND NOT ENABLE_WAFFLE)
     install (TARGETS eglretrace RUNTIME DESTINATION bin) 
 endif ()
 
+find_package(PkgConfig)
+pkg_search_module(GBM gbm)
+
+if (ENABLE_EGL AND GBM_FOUND AND NOT WIN32 AND NOT APPLE AND NOT ENABLE_WAFFLE)
+    add_executable (eglretrace_gbm
+        glws_egl_drm.cpp
+    )
+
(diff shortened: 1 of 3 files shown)


linkmauve:
Pushed a second version, the only differences being that snapshots do work (I had forgotten to call eglMakeCurrent() after a resize), and the underlying buffer is now ARGB instead of XRGB, to give back alpha values.

evelikov:
Afaics if you use the waffle backend (+gbm) you'll already get support for render nodes. You might need to feed the node name to display_connect() or via WAFFLE_GBM_DEVICE [1].

[1] https://github.com/waffle-gl/waffle/commit/cef67496e379b660601f64ab2ddf7cf06307b8aa

jrfonseca:
I think @evelikov has a point.

For major desktop APIs (GLX/CGL/WGL) I'll still want to support them directly for my own convenience, but there's no point in duplicating all EGL variants in apitrace when waffle abstracts them away.  

And with commit 661f2cfe53303dd60676e7fd0b25440be80d6644 it's now possible to choose the WAFFLE_PLATFORM on Linux.

Issue #370, opened 2015-08-12
std::regex_error with GCC 4.7 / 4.8


farmboy0:
The error message is:
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
Aborted

I bisected it to commit 36b945cb98a4e9009d57b8c20a720fc1a5905452

This happens for qapitrace as well the cli tools.

jrfonseca:
I've been using this on Linux without problems. What linux/ gcc / libstdc++ you have?

farmboy0:
I am using GCC 4.8.4 on Gentoo

jrfonseca:
It looks like std::regex was only properly supported on GCC 4.9:

   http://stackoverflow.com/questions/12530406/is-gcc-4-7-and-gcc-4-8-buggy-about-regular-expressions

I'll to think a bit more how to handle this. On one hand regex are really handy. But it's not clear how much important is GCC 4.8 compatability.

In the meanwhile, if you can upgrade to GCC 4.9 that would be the idea.

farmboy0:
why do you need those regexes on linux anyway?

jrfonseca:
So I can dump Windows traces on Linux. Furthermore I plan to use regexes for glDraw* calls too.

Regardless of the API, maintainging these table by hand is not sustainable, as every new Direct3D version or OpenGL requires manually adding more entries.  While a regex is future proof.

farmboy0:
Then I guess you will need to bump the gcc check to 4.9.

farmboy0:
I have rebuilt apitrace with GCC 4.9.3 and it works again.

jrfonseca:
> Then I guess you will need to bump the gcc check to 4.9.

Yes, I'll do that for now.

But I'm not sure if making GCC 4.9 a hard requirement is too aggressive, and if I should make it a soft requirement somehow, even if functionality is poorer.

farmboy0:
how can you do that if you want to have the regexes?

mtschoen:
Couldn't we just wrap the regex calls in one of these?

https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

jrfonseca:
> Couldn't we just wrap the regex calls in one of these?

Not unless we loose functionality, or add back the non-regex code to do it.  Neither ideal.

jrfonseca:
> how can you do that if you want to have the regexes?

[...]

PR #382, opened 2015-08-26
Improving profiling capabilities of glretrace.

trtt:
This was done as a part of my GSoC summer project.
Here is somewhat detailed description of the changes:
https://github.com/trtt/apitrace/wiki/Description

I am sure there is a lot to be improved. Looking forward to receiving some feedback.

### DIFF
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -82,6 +82,7 @@ add_library (glretrace_common STATIC
     metric_writer.cpp
     metric_backend_amd_perfmon.cpp
     metric_backend_intel_perfquery.cpp
+    metric_backend_opengl.cpp
 )
 add_dependencies (glretrace_common glproc)
 target_link_libraries (glretrace_common
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -78,6 +78,7 @@ add_library (glretrace_common STATIC
     glstate_params.cpp
     glstate_shaders.cpp
     glws.cpp
+    metric_helper.cpp
 )
 add_dependencies (glretrace_common glproc)
 target_link_libraries (glretrace_common
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -80,6 +80,7 @@ add_library (glretrace_common STATIC
     glws.cpp
     metric_helper.cpp
     metric_writer.cpp
+    metric_backend_amd_perfmon.cpp
 )
 add_dependencies (glretrace_common glproc)
 target_link_libraries (glretrace_common
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -81,6 +81,7 @@ add_library (glretrace_common STATIC
     metric_helper.cpp
     metric_writer.cpp
     metric_backend_amd_perfmon.cpp
+    metric_backend_intel_perfquery.cpp
 )
 add_dependencies (glretrace_common glproc)
 target_link_libraries (glretrace_common
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -79,6 +79,7 @@ add_library (glretrace_common STATIC
     glstate_shaders.cpp
     glws.cpp
     metric_helper.cpp
+    metric_writer.cpp
 )
 add_dependencies (glretrace_common glproc)
 target_link_libraries (glretrace_common
--- a/retrace/glretrace.hpp
+++ b/retrace/glretrace.hpp
@@ -143,6 +143,8 @@ MetricBackend* getBackend(std::string backendName);
 
 bool isLastPass();
 
+void listMetricsCLI();
+
 void enableMetricsFromCLI(const char* metrics, QueryBoundary pollingRule);
 
 GLenum
--- a/retrace/glretrace.hpp
+++ b/retrace/glretrace.hpp
@@ -72,6 +72,11 @@ struct Context {
     }
 };
 
+extern bool metricBackendsSetup;
+extern bool profilingBoundaries[QUERY_BOUNDARY_LIST_END];
+extern std::vector<MetricBackend*> metricBackends;
+extern MetricBackend* curMetricBackend;
+
 extern glprofile::Profile defaultProfile;
 
 extern bool supportsARBShaderObjects;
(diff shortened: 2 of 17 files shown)


jrfonseca:
Was this pull-request closed by mistake, or will you submit a new one?

trtt:
Did not know that apitrace made it back to gcc 4.8.
I closed it as it did not compile and I was a bit surprised.
I made a commit to support some features in 4.8, but it is still not enough.
std::regex code does not link. I guess there should be an alternative found.
I think I can reopen it though.

jrfonseca:
> Did not know that apitrace made it back to gcc 4.8.

Ah, yes.  I too tried to use std::regex but that cause problems for several people/platforms due the gcc 4.9 requirement.  See issue #370.

I'm not sure exactly when gcc 4.9 will become widely available.

If you can somehow avoid std::regex, that would be the ideal.

trtt:
Replaced std::regex code.

jrfonseca:
@trtt another comment: given the complexity I think it would be worth trying to get some code coverage of it in https://github.com/apitrace/apitrace-tests

Have you any thoughts or ideas on that?

Unfortunately the apitrace-tests don't cover profiling yet, but in fact even the basic profiling we have today run into a bunch of bugs (mostly due to extensions checking) so I've been meaning to get some coverage there too.

mupuf:
Thanks for the review José!

jrfonseca:
@trtt, I started to merge, but the branch fails to build on Mac due to some incompatibility with libc++.

It's also possible to reproducing the build failure on linux, by doing e.g., on Ubuntu:

    sudo aptitude install clang libc++-dev
    cmake -H. -Bbuild/clang -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS=-stdlib=libc++
    cmake --build build/clang --use-stderr
    [...]
    In file included from ../../retrace/metric_backend_amd_perfmon.cpp:26:
    In file included from ../../retrace/metric_backend_amd_perfmon.hpp:35:
    In file included from ../../retrace/glretrace.hpp:31:
    In file included from ../../retrace/metric_writer.hpp:3:
    In file included from /usr/include/c++/v1/queue:169:
    /usr/include/c++/v1/deque:294:67: error: no viable conversion from 'const std::__1::deque<unsigned int *, std::__1::allocator<unsigned int *> >' to 'unsigned int *const'
        _LIBCPP_INLINE_VISIBILITY reference operator*() const {return *__ptr_;}
                                                                      ^~~~~~~
    [...]

I can't make much sense of the errors.

The funny thing is that "metric_writer.hpp" is included from "glretrace.hpp", but there are no compilation failures for files that include "metric_writer.hpp" through "glretrace.hpp".
So I suspect the problem is that "metric_writer.hpp" is missing some includes.

But it's possible the problem only happens on certain files where the templates/classes get actually instantiated.

I'm afraid that Mac OS is too important, so this must be addressed before the branch can be merged.

Can you help?

trtt:
Sorry, was a bit busy, this pull request completely escaped my mind.

I made a small fix, now it works with libc++. When it comes to things like allocators, error messages are really misleading, I guess. Also, the fact that it worked with libstdc is misleading as well.

I think there are some merge conflicts with more recent commits to the project. Tell me if I need to fix that.

trtt:
@jrfonseca 


> The problem with multiple passes within a single glretrace process is that it's not bullet proof: we maintain several things in globals, and they not all get reset after the replaying. Even if we cleanup all globals, there's still the risk that some global state in the graphics APIs we use cause behavior change.
> 
> Rather than doing multiple passes inside glretrace, it might be more robust to run glretrace process multiple times (you could even have glretrace spawn itself for each pass.)

I see the problem. I think it is a bit difficult to make it work like that. To do it in multiple runs (process runs) some kind of shared memory is required (for keeping backend states and data). Something like Boost::Interprocess, so that it also works consistently on all platforms. Do you know what can be used here in Apitrace?



> given the complexity I think it would be worth trying to get some code coverage of it in https://github.com/apitrace/apitrace-tests
> 
> Have you any thoughts or ideas on that?

It's difficult for me to think of something here. I don't know what kind of tests fits the profiling. The set of metrics and everything is very hardware dependent. With --pcpu,--pgpu it was more or less consistent.
Perhaps, I'm missing something. Maybe you have some examples of what could approximately be done?

jrfonseca:
As explained in https://github.com/trtt/apitrace/commit/136411b993a1631260c553de8bc9a9102398bca6#commitcomment-15264358 I don't feel I can maintain multiple passes going forward, so I can't merge this as is.

> To do it in multiple runs (process runs) some kind of shared memory is required (for keeping backend states and data). Something like Boost::Interprocess, so that it also works consistently on all platforms. Do you know what can be used here in Apitrace?

No shared memory is necessary -- each pass can ran independently, and then data merged separately.

But honestly, I think it should be OK to have the end-user to select a manageable subset of counters, instead of trying to give the illusion we can cope with an arbitrary number of counters at the expense of high complexity in apitrace.   It would be already a big improvement.


I tried to yank out the multiple pass support myself, but it's tightly rooted all over the place.  I'm afraid it would take more time I can afford to.

As it stands I don't plan to do any more work on this pull request.  If there's still somebody willing to do the work I'll merge it.  Otherwise I'll have to close it.


I'm sorry things turned this way.  But to be perfectly honest, this is the likely outcome when such complex features are developed with no attempt to keep maintainer in the loop.  I strongly wished this GSoC would be a success so there would be more in the future but I was deceiving myself.  There were already several pull requests in the past where I sort of closed a blind eye then regretted, because code/feature just silently rots, then I get bug reports.  So I don't want to do that anymore.

jrfonseca:
@trtt, @mupuf, I finally merged this.  Thanks for this large contribution and my apologies for taking this long.

@trtt, one last request: could you please add copyright headers to all new files -- use one of the existing headers and replace the author with your name (or organization if it applies).

And if any of you still has time to put on this feature, it would be very useful to have some sort of test coverage (some script in apitrace-test).  Because this is a large amount of code, so without any test coverage it can easily regress when nobody is looking at it, which would be a pity.

Particularly the fact that this functionality requires multiple pass seems a brittle point (as no other feature in apitrace relies on it), so I can easily imagine breaking due to some invalid assumptions done in globals.  So at very least it might be worth having a test that ensures that multi-pass doesn't regress (maybe adding a hidden option to glretrace to force multiple passes, to stress this code independently of profiling backend).

mupuf:
Thanks a lot Jose!

You are right about the tests, I will try to think of how to test the code since every machine has a different set of counters exported and will return different values due to internal architecture differences (unless it is measuring stuff related to the GL pipeline). We can try to check for consistency of the results at least. I am happy to help trtt on this (thanks a lot of sticking around trtt!)!

I guess we also need to make this work easier to use. Janesma worked on improving the GUI for analysing performance, it would be nice to see if we can hook it up to this new feature. If not, I could mentor trtt or another student for the GSoC 2016. This time, I will really ask for incremental updates to make sure everything improves in the right direction and won't be a nightmare to get merged :)

PR #383, opened 2015-08-27
update the path of egltrace.so

joshua5201:
correct the path of compiled egltrace.so


joshua5201:
I'm merging other changes into one PR.

PR #384, opened 2015-08-28
Change dalvik.markdown, letting the instructions work

joshua5201:
the original one doesn't work for me, the egltrace.so is in different path and I need to LD_PRELOAD libgnustl.so.


jrfonseca:
Pushed. Thanks.

PR #387, opened 2015-09-11
Fix Waffle detection on non-Android

linkmauve:
On a standard ArchLinux with aur/waffle installed, the current FindWaffle.cmake can’t find the library because it is in /usr/include/waffle-1/ instead of directly in /usr/include/ like it is expecting.

Instead of hardcoding this path, I added pkg-config support so it will work wherever the user has Waffle installed, as long as they provide a .pc file.


jrfonseca:
I'm OK with the patch, but I believe that noawadays Waffle ships .cmake files so this can be done even easier.

I don't recall the exact details, but I believe that piglit's CMakeLists was update to import waffle like that at one point.

jrfonseca:
This didn't work for me -- I had to rename `PkgConfig_FOUND` to `PKG_CONFIG_FOUND`

linkmauve:
Here is a new version using the FindWaffle.cmake script provided by Waffle itself.

jrfonseca:
> Here is a new version using the FindWaffle.cmake script provided by Waffle itself.

@linkmauve Thanks. I already pushed this, plus a few more changes.  Could you please rebase?

PR #391, opened 2015-09-21
Add parallel snapshotting in retrace

linkmauve:
The first commit is just a cleanup.

The second one adds a new option, -t/--snapshot-threaded, enabling threaded encoding of the snapshots into PNG files, using as many threads as there are cores on the machine.

### DIFF
--- a/common/os_thread.hpp
+++ b/common/os_thread.hpp
@@ -73,6 +73,7 @@ namespace os {
 #else
 #  include <pthread.h>
 #endif
+#include <functional>
 
 
 namespace os {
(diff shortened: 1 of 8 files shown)


jrfonseca:
I think this is an excellent idea.

Yes, I'm afraid we can't depend on std::thread yet. If you can get it to work with os::thread I'm happy to merge.  I can help with the Windows side.

Why did you need to modify os::thread creation?

linkmauve:
I needed it to accept different number of arguments, one for the relay race and zero for this PR, so I thought converting that into a template was the easiest way to achieve it, but I haven’t been able to get it to work correctly, it seems to try to call operator() while we want a constructor.

jrfonseca:
I see. Yes, std::thread allows variadic arguments.  And properly mimicing probably implies templates with variadic arguments. But it's probably not worth to dwell some much time in faithfully matching std::thread, as sooner or later we'll migrate to true std::thread. Instead, I'd recommending adding a dummy integer parameter to this PR's thread (with default value of zero).

jrfonseca:
It took me quite a while, but I finally managed to implementing the necessary Windows bits in os_thread.hpp, both for Vista+ and XP.

Unfortunately, even after all this time, C++11 threading is not widely available on Windows (MSVC has, but not for XP, and  MinGW only have on top of POSIX thread emulation library I rather not depend upon.)

Anyway, this has finally been commited as c7f3b9b6855d5bc913b396b9da6f841d5c58730c .  Thanks for your contribution and patience.

PR #392, opened 2015-09-24
Added images (glBindImage) to displayed surfaces

schulmar:
First time contribution so please bear with me.

## Summary ##
I added dumping and display of textures (that are bound via glBindImage) as "Images" under Current State->Surfaces.

## Motivation ##
I am working with many compute shaders and they're more likely to access textures via image units than via samplers.

## TODOs ##
I am sure I made many mistakes that you have to review but it works at least on my local machine.

As glBindImage is from a more recent GL version than vanilla textures it might be necessary to test if it is available before dumping. 
I have no experience in how to do that within apitrace code so I hope for some helpful reviews.

I am relying on GL_ARB_direct_state_access to get the type of the texture from the texture ID. There are less efficient but more compatible ways to do that but I have gone for this one because it has been available for some time now.

### DIFF
--- a/gui/apitracecall.cpp
+++ b/gui/apitracecall.cpp
@@ -436,6 +436,33 @@ ApiTraceState::ApiTraceState()
 {
 }
 
+ApiTexture getTextureFrom(QVariantMap const &image, QString label) {
+    QSize size(image[QLatin1String("__width__")].toInt(),
+               image[QLatin1String("__height__")].toInt());
+    QString cls = image[QLatin1String("__class__")].toString();
+    int depth =
+        image[QLatin1String("__depth__")].toInt();
+    QString formatName =
+        image[QLatin1String("__format__")].toString();
+
+    QByteArray dataArray =
+        image[QLatin1String("__data__")].toByteArray();
+
+    QString userLabel =
+        image[QLatin1String("__label__")].toString();
+    if (!userLabel.isEmpty()) {
+        label += QString(", \"%1\"").arg(userLabel);
+    }
+
+    ApiTexture tex;
+    tex.setSize(size);
+    tex.setDepth(depth);
+    tex.setFormatName(formatName);
+    tex.setLabel(label);
+    tex.setData(dataArray);
+    return tex;
+}
+
 ApiTraceState::ApiTraceState(const QVariantMap &parsedJson)
 {
     m_parameters = parsedJson[QLatin1String("parameters")].toMap();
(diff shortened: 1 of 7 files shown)


jrfonseca:
@schulmar , thanks for your patch.  A few comments:

- I think it's simpler and better not to modify the UI. Merely dump the images inside the "textures" JSON object

- it's fine to rely on `GL_ARB_direct_state_access` but dumpImages should check whether `GL_ARB_shader_image_load_store` and  `GL_ARB_direct_state_access` are support, and do nothing when it's not.  Otherwise it might cause GL errors or crashes in implementations that don't support those extesnions.  This can be done by adding flags for these extensions to the glstate::Context object.

schulmar:
@jrfonseca I changed the code according to your comments. 

Currently I have no time for testing whether the extension checking works correctly in the negative case, maybe someone else can confirm that.

There seems to be a problem with merging the images into the textures in the JSON dump:
No "normal" textures get displayed. I believe that is caused by dumpTextures and dumpImages both  creating a "textures" member where the latter overrides the first.

There are several possible solutions to this:
- don't reuse the textures dump member (my initial solution)
- put them in the same textures dump member, which means either
 - move begin/endMember("textures") from dumpTextures and dumpImages into the calling code, wrapping both calls
 - remove begin/endMember("textures") from dumpImages and call it from dumpTextures or vice versa

Maybe you have some preference here? 
Could you explain why you think it is better to dump the images into the textures JSON object? 
I understand that the handling is basically the same, but I find it cleaner to have them separated (especially from the UX viewpoint).

schulmar:
@jrfonseca Since you did not answer I decided to go with the solution that fit your last comments and was the best in my opinion:

I added a parameter to the dumpImages function that allows to turn on/off whether the bound images should be dumped into an "images" member of the dump. Now I call this function from within dumpTextures, dumping into the surrounding "textures" member.

schulmar:
@jrfonseca Since more time has passed without any word from your (or any other contributors) side I want to know if there is a realistic chance of this making it into the master branch? If there are more things necessary to get this to "pass" then please tell me what to do. 

I would like to contribute further features but it does not make sense to go on when even this small one will not pass.

jrfonseca:
@schulmar, thanks for the updates, I'm not aware of any outstanding issue. I've been meaning to merge it but I haven't yet had the time.

Lately I have had very little time for apitrace development both on and off my day job, it's still crucial so I need to be thorough to ensure nothing breaks. The end result is a backlog of patches/branches to merge, not to mention lots of half-finished experimental branches of my own...

jrfonseca:
Finally merged.

I squashed the series into two patches:
* 34b56c7727184a9b4c746ecd289c86a85ca2d5b0
* 638ec448619fda80fcb439b1747af62169d05548

Thanks for your patience.  I hope the delay doesn't demotivate you from further contributions.  I'll try to do better next time.

schulmar:
Sorry if my post seemed aggressive. 
I have no problem with waiting but the lack of communication was a bit problematic.

I definitely plan to contribute further but I don't know when I'll have time for that.

PR #417, opened 2016-01-20
scripts: output same subject leak sorted by generate call no

comicfans:
misc improve on leak script


jrfonseca:
I merged this into leaks and pushed into master.

PR #418, opened 2016-01-25
docs: add description to leak trace function

comicfans:
this is description for leak trace usage, but I'm not sure this is correct. I think current python implementation didn't trace object 'per thread' (although 'per-context' is exactly correct) .


jrfonseca:
I pushed with a few tweaks.

PR #422, opened 2016-01-30
Add cmake option for lssp linking in mingw

EoD:
Some cross-compiled compilers have ssp built in, no need to link it forcefully.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,8 @@ option (ENABLE_WAFFLE "Enable WAFFLE support." OFF)
 
 option (ENABLE_FRAME_POINTER "Disable frame pointer omission" ON)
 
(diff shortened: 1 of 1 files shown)


jrfonseca:
> Some cross-compiled compilers have ssp built in, no need to link it forcefully.

Could you provide more detail? What exactly do you mean by `built in`, and where can I get such crosscompiler?

Instead of an option, maybe we can do a quick configure time check (using cmake `try_compile`  command or CheckCXXSourceCompiles/CheckCXXSymbolExists modules.), and automatically do the right thing without requiring the user to manually set the option.

EoD:
I think my crosscompile setup is broken and I misunderstood the solution. I will withdraw this PR in its current form, if you don't object.

jrfonseca:
Sure.

Issue #350, opened 2015-06-10
apitrace: error: couldn't load /System/Library/Frameworks/OpenGL.framework/OpenGL


qiweiye:
For my machine, it's Mac OSX 10.10.

 build the apitrace locally after installing the cmake & qt.

the build process is successfully. But when I try to use it, get below error:

qiweiye-trash:build qiweiye$ ./apitrace trace -a gl /Applications/Klondike\ Forever.app/Contents/MacOS/Klondike\ Forever 
apitrace: loaded into /Applications/Klondike Forever.app/Contents/MacOS/Klondike Forever
apitrace: tracing to /Users/qiweiye/Library/Containers/com.solitaireforever.Klondike-Forever/Data/Klondike Forever.trace
apitrace: error: couldn't load /System/Library/Frameworks/OpenGL.framework/OpenGL

jrfonseca:
One of the calls in https://github.com/apitrace/apitrace/blob/master/dispatch/glproc_gl.cpp#L105 is failing.

Could you try figure which?

Either adding some printfs, or using the debugger as described in https://github.com/apitrace/apitrace/blob/master/docs/BUGS.markdown#mac-os-x

jrfonseca:
The problem is described in [here](https://developer.apple.com/library/mac/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxInDepth/AppSandboxInDepth.html#//apple_ref/doc/uid/TP40011183-CH3-SW17):

    Note: The /tmp directory is not accessible from sandboxed apps. You must use the NSTemporaryDirectory function to obtain a temporary location for your app’s temporary files.

But `NSTemporaryDirectory` is an Objective-C function, so we'd need to write a C wrapper... or avoid the symlink hack altogether somehow...

PR #425, opened 2016-02-13
Fixed loading of OpenGL.framework in snaboxed environment.

rolan-reznik:
Hello, I have fixed issue #350. Any comments and feedback are welcome.


jrfonseca:
Thanks for doing this.  I just cleaned up code a bit and pushed it.

PR #437, opened 2016-03-23
List Shader Storage Buffer Blocks

schulmar:
Add a new tab for Shader Storage Buffer Blocks from all shaders.

This is the prerequisite for a new viewer that will offer to inspect a buffer via the structure of a specific shader storage buffer block.

The viewer is already being worked on but I want to wait for this to make it into master before relying on it. The viewer would need the contents of buffers, which would have to be exported as well. We would need to discuss how to export them as they might be several GiB in size.

### DIFF
--- a/retrace/glstate_shaders.cpp
+++ b/retrace/glstate_shaders.cpp
@@ -1138,6 +1138,28 @@ void dumpShadersStorageBufferBlocks(StateWriter &writer, Context &context,
                 "GL_BUFFER_DATA_SIZE",
                 getProgramResourcei(program, GL_SHADER_STORAGE_BLOCK,
                                     ssbbResourceIndex, GL_BUFFER_DATA_SIZE));
+
+            auto outputIfReferenced = [&writer,ssbbResourceIndex, program](
+                const char *propertyName, GLenum property) {
+                const auto value =
+                    getProgramResourcei(program, GL_SHADER_STORAGE_BLOCK,
+                                        ssbbResourceIndex, property);
+                if (value) {
+                    writer.writeBoolMember(propertyName, true);
+                }
+            };
+#ifdef OUTPUT_IF_REFERENCED
+#error "OUTPUT_IF_REFERENCED should not be defined here!"
+#endif
+#define OUTPUT_IF_REFERENCED(NAME) outputIfReferenced(#NAME, NAME)
+
+            OUTPUT_IF_REFERENCED(GL_REFERENCED_BY_VERTEX_SHADER);
+            OUTPUT_IF_REFERENCED(GL_REFERENCED_BY_TESS_CONTROL_SHADER);
+            OUTPUT_IF_REFERENCED(GL_REFERENCED_BY_TESS_EVALUATION_SHADER);
+            OUTPUT_IF_REFERENCED(GL_REFERENCED_BY_GEOMETRY_SHADER);
+            OUTPUT_IF_REFERENCED(GL_REFERENCED_BY_FRAGMENT_SHADER);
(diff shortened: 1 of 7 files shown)


retrace/glstate_shaders.cpp
```diff
@@ -1138,6 +1138,28 @@ void dumpShadersStorageBufferBlocks(StateWriter &writer, Context &context,
...
+                    getProgramResourcei(program, GL_SHADER_STORAGE_BLOCK,
+                                        ssbbResourceIndex, property);
+                if (value) {
+                    writer.writeBoolMember(propertyName, true);
+                }
+            };
+#ifdef OUTPUT_IF_REFERENCED
+#error "OUTPUT_IF_REFERENCED should not be defined here!"
+#endif
+#define OUTPUT_IF_REFERENCED(NAME) outputIfReferenced(#NAME, NAME)
```
jrfonseca: Instead of the macro, please just use `enumToString` helper function.

retrace/glstate_shaders.cpp
```diff
@@ -1138,6 +1138,28 @@ void dumpShadersStorageBufferBlocks(StateWriter &writer, Context &context,
                 "GL_BUFFER_DATA_SIZE",
                 getProgramResourcei(program, GL_SHADER_STORAGE_BLOCK,
                                     ssbbResourceIndex, GL_BUFFER_DATA_SIZE));
+
+            auto outputIfReferenced = [&writer,ssbbResourceIndex, program](
```
jrfonseca: Instead of the lambda, please just use a static const table of GLenum, and a for cycle.

retrace/glstate_shaders.cpp
```diff
@@ -1049,6 +1049,136 @@ dumpShadersUniforms(StateWriter &writer, Context &context)
...
+                writer.beginObject();
+
+#define DUMP_BUFFER_VARIABLE_INT(NAME)                                         \
+    do {                                                                       \
+        int value = getProgramResourcei(program, GL_BUFFER_VARIABLE,           \
+                                        variableIndex, NAME);                  \
+        writer.writeIntMember(#NAME, value);                                   \
+    } while (false)
+
+                DUMP_BUFFER_VARIABLE_INT(GL_TYPE);
```
jrfonseca: Like my other comment, please use GLenum table, plus enumToString to obtain C strings.

jrfonseca:
The changes look good overall.  I just made few cosmetic requests as noted, with the intention of making the code a bit simpler and easier to read. 

> We would need to discuss how to export them as they might be several GiB in size.

For starters we could truncate when they are too big.

Long term, we might need to start storing state dumps on disk.

schulmar:
I applied the changes you suggested and added some more information to the value field.

Regarding the size problem I would say that truncating does not seem like the solution.
My idea would be to not export the buffers per default and to add an option to export specific buffers on demand. This way the buffers could be reloaded when the user wants to inspect it.

A similar problem can happen with textures. I have programs with textures that reach multiple GiB of size. However, for textures a small thumbnail version should be exported to be visible in the surfaces list. Should I open a separate issue to discuss this further?

jrfonseca:
Thanks for the update.

I only suggested truncating as a stopgap solution (to avoid qapitrace crashing due to OOM, particularly on 32bits where address space is scarse.)  I agree we should figure something better out.

> Should I open a separate issue to discuss this further?

Yes, please.  This way we can close this one when I merge your pull request.

schulmar:
>>    Should I open a separate issue to discuss this further?
>Yes, please. This way we can close this one when I merge your pull request.
I opened Issue #438 for this.

jrfonseca:
I've squashed your commits and pushed them.  Thanks.

PR #448, opened 2016-04-16
gui: Add checkboxes for alpha/opaque in surfaces view (fixes #374)

### LINKED ISSUE #374: GUI: "opaque" and "alpha" as option for surface tab
The image viewer has the opaque option. It would be great if the surfaces tab also had an opaque option.
Our application renders the color framebuffer with alpha = 0 most of the time.
Unfortunaly this means that the framebuffer preview in the surfaces tab stays "empty". I first have to double click it to get into the image viewer, then use the "opaque" option to see the current framebuffer content.
If the surface tab would have the opaque option which would toggle opaqueness of the preview images that'd be great. Other people could probably use an alpha option too, so it probably makes sense to add both at once.

schulmar:
- pass alpha/opaque settings through to image viewer
- removed caching of thumbnails in ApiSurface instances
  - updating thumbnails (due to changed alpha/opaque settings) did not
mesh with constness
  - ApiSurfaces should be unaware of thumbnails -> caching should happen
e.g. in SurfacesView

### DIFF
--- a/gui/apitracemodel.cpp
+++ b/gui/apitracemodel.cpp
@@ -75,7 +75,7 @@ QVariant ApiTraceModel::data(const QModelIndex &index, int role) const
                         ", %1kb";
 
                 ApiFramebuffer fbo = frame->state()->colorBuffer();
-                QImage thumb = fbo.thumb();
+                QImage thumb = fbo.calculateThumbnail(false, false);
                 if (!thumb.isNull()) {
                     QByteArray ba;
                     QBuffer buffer(&ba);
--- a/gui/apisurface.cpp
+++ b/gui/apisurface.cpp
@@ -2,6 +2,7 @@
 #include "thumbnail.h"
 
 #include <sstream>
+#include <memory>
 
 #include <QDebug>
 #include <QSysInfo>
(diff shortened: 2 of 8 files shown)


gui/apitracemodel.cpp
```diff
@@ -75,7 +75,7 @@ QVariant ApiTraceModel::data(const QModelIndex &index, int role) const
                         ", %1kb";
 
                 ApiFramebuffer fbo = frame->state()->colorBuffer();
-                QImage thumb = fbo.thumb();
+                QImage thumb = fbo.calculateThumbnail(false, false);
```
schulmar: There also is a frame->thumbnail which comes from gathering the thumbnails from retrace. 

This smells like duplicated functionality but I am not sure.

schulmar:
As you can see, I removed the caching. I simply had no trace where it would be necessary. If you have one, I'd like to use it for testing where the caching should be reintroduced.

It might be possible to store the general thumbnail for `alpha=false`, `opaque=false`, `lowerValue=0`, and `upperValue=1` and then use QImage/QPixmap transformations and  to react to changes in `opaque` and `alpha`.

jrfonseca:
Looks good AFAICT.

I've only did a couple cosmetic tweaks (change the case of the checkbox, and remove the DefaultLowerValue/DefaultUpperValue since we'll never change it, and 0/1 seems pretty self explanatory), and pushed.

I think that removing the thumbnail cache in state viewer is fine.

I think it's ok to leave thumbnails for frame/draw calls alone for now.  Specially because we don't keep the original images.

We could eventually have the Opaque/Alpha setting in the menu instead of surfaces tab, and have it apply globally to all thumbnails.  That would imply to do the transformation thumbnail -> thumbnail for the call list thumbnails.

Issue #427, opened 2016-02-24
Crash retracing file in qubjson


gerboland:
Tracing this file: https://dl.dropboxusercontent.com/u/85539674/qmlscene.trace I get a crash (double-clicked last frame in qapitrace):

Running: "eglretrace" "-D" "54236" "--dump-format" "ubjson" "qmlscene.trace"
../gui/qubjson.cpp:138: int readSize(QDataStream&, ubjson::Marker): Unimplemented code.
ASSERT: "0" in file ../gui/qubjson.cpp, line 284

Works with 2d5b6b107d77a31935aa4fe53ba67a8574372672
I can bisect if you need

jlegg0:
I can reproduce this crash with a short trace (30 calls) at commit 53e7386ff756f44334d8e3648b3420e407d5eb14. I've attached the trace, but I added .txt onto the end of the filename as this bug tracker complains otherwise.
[apitrace-427-minimal.trace.txt](https://github.com/apitrace/apitrace/files/372017/apitrace-427-minimal.trace.txt)

jlegg0:
I bisected this to commit f6305d3d105376de5306f6d0e70ff666df38d2e0 using my trace.

jlegg0:
Pull request #473 has a fix for the crash. This skips dumping of vertex attributes in buffers without GL_MAP_WRITE_BIT in their storage flags however. Maybe the recorded storage flags could be altered by glretrace to allow dumping?

jlegg0:
Hm. I've found another way of failing that assert that is not fixed by #473. I'll investigate.

jlegg0:
I''ve added a second commit to #473 which fixes the other problem I found.

jrfonseca:
> Maybe the recorded storage flags could be altered by glretrace to allow dumping?

@jlegg0 Yes, that's a possibility -- it might introduce divergences from the original app, but we've done that sort of stuff before.  Would glGetBufferData work however? That would avoid interfering.

jlegg0:
Good call. glGetBufferSubData is unaffected by the storage flags, so it would work in some cases where mapping would fail without interfering with storage flags. However, glGetBufferSubData will fail if the buffer was already mapped (as long as it wasn't persistently mapped). The current mapping path will work when the part of the buffer that will be used is already mapped and read access was enabled, but not in other cases where the buffer was already mapped. To work in all situations, glretrace could temporarily unmap buffers when they are mapped non-persistently and use glGetBufferSubData.

[...]

PR #473, opened 2016-07-20
Fixes for #427

jlegg0:
Fixes https://github.com/apitrace/apitrace/issues/427


jlegg0:
Thanks for merging these.

PR #476, opened 2016-07-31
d3dretrace: Add profiling support

siro20:
Add profiling support for D3D9. It's based on glretrace code and
Tiziano Bacocco initial attempt.

Crosscompiled using mingw64 and tested on WINE.


jrfonseca:
@siro20, @iiv3, I tried to merge this, but I'm afraid this still needs additional work.

When I tried to run `d3dretrace --pgpu ...` on Windows it immediately hung.

Looking at the source code, snippets like

    while(query.ids[GPU_START]->GetData(&gpuStart, sizeof(UINT64), 0) != S_OK) {}

look wrong in a few ways:
- One should use `D3DGETDATA_FLUSH`, to ensure the commands are actually flushed
- One should check for `S_FALSE`, to avoid deadlocking on errors

See https://docs.microsoft.com/en-us/windows/desktop/direct3d9/queries for more info.

There are some matters of code style, but that I can gloss over for now.  But this needs to work minimally well not only on WINE but on native Windows before I can merge.

It's not too hard to fix it myself, but I don't think I'll have the time now, as I'm looking for the low hanging fruit PRs I can merge without too much manual intervention.

mirh:
Any more love for this?

PR #478, opened 2016-08-13
d3dretrace: Fix a crash

siro20:
ATI1N and ATI2N are compressed formats, like DXTn.
As compressed formats are smaller in size, make sure to calculate correct buffer size.
Fixes a crash while replaying Unigine Valley Benchmark.

### DIFF
--- a/helpers/d3d9size.hpp
+++ b/helpers/d3d9size.hpp
@@ -144,10 +144,12 @@ _getFormatSize(D3DFORMAT Format, size_t & BlockSize, UINT & BlockWidth, UINT & B
         break;
 
     case D3DFMT_DXT1:
+    case D3DFMT_ATI1N:
         BlockHeight = BlockWidth = 4;
         BlockSize = 64;
         break;
 
+    case D3DFMT_ATI2N:
     case D3DFMT_DXT2:
     case D3DFMT_DXT3:
     case D3DFMT_DXT4:
@@ -159,16 +161,13 @@ _getFormatSize(D3DFORMAT Format, size_t & BlockSize, UINT & BlockWidth, UINT & B
     case D3DFMT_NV12:
     case D3DFMT_YV12:
         // Planar YUV
-    case D3DFMT_ATI1N:
-    case D3DFMT_ATI2N:
-        /*
-         * Because these are unsupported formats, RowPitch is not set to the
-         * number of bytes between row of blocks, but instead in such way that
-         * Height * RowPitch will match the expected size.
-         */
-        BlockWidth = 0;
-        BlockSize = 0;
-        break;
+        /* Because these are unsupported formats, RowPitch is not set to the
+        * number of bytes between row of blocks, but instead in such way that
+        * Height * RowPitch will match the expected size.
+        */
+       BlockWidth = 0;
+       BlockSize = 0;
+       break;
 
     case D3DFMT_UNKNOWN:
     case D3DFMT_VERTEXDATA:


jrfonseca:
Are you sure about this one?

We have several traces with ATI1N/ATI2N taken with NVIDIA and VMware D3D9 drivers and they are working correctly.

Was this with WINE?  Could it be that WINE is doing something weird?

Because these formats are very peculiar.  They are not like DXTn formats at all.  This is because the Direct3D runtime knows about DXTn have 4x4 block, and adjusts things accordingly.  But ATI1N/ATI2N are unnofficial formats, so the Direct3D Runtime doesn't know they have a 4x4 block.   The D3D9 drivers need to fiddle with the `PrivateFormatBitCount` of [FORMATOP structure](https://msdn.microsoft.com/en-us/library/windows/hardware/ff566438.aspx), and fiddle with the pitch, so that Direct3D memcpy'es the _right_ ammount of pixels. 

The rule is that one should use `height * pitch`, and that the driver is adjusting pitch to be the right value.

siro20:
Th trace was tested on WINE an Gallium Nine. I'm not aware of such a rule. Locking ATIx would give a pitch that's smaller than width, as the height is fixed?

jrfonseca:
My understanding is that the D3D9 driver/implementatino should return a `PrivateFormatBitCount == 8`, that ie, 8 bits per pixel.  Which is about the right density, as 128 bits for every 4x4 pixels is 8 BPP.

But because D3D runtime copies height x pitch, this means it ends up transfering more pixels than it should.

See `dx9_tex_formats_ati?n.trace` traces in https://people.freedesktop.org/~jrfonseca/traces/ 

Anyway, I'm not 100% sure myself.  I'll try to double check.  But this is my current understanding.

jrfonseca:
If you grep Wine source code for `WINED3DFMT_FLAG_BROKEN_PITCH` you'll see that Wine does precisely what I describe.

Nine also seems to do the right thing.  So I don't understand why this patch makes a difference.

Could you please provide the trace of Unigine Valley Benchmark, without this patch (ie, the one that caused the crash)?

siro20:
After checking sources I can confirm that this patch applies to ATI1 only, as ATI2 has the correct buffer size by coincident.

For ATI1 4x4 pixels take only 64bits, ATI2 takes 128bits.

WINE and NINE knows about ATIn and advertise it as uncompressed 8bpp.
I've had a look at the dx9_tex_formats_ati1n.trace and it does the same.

Without this patch:
A texture with dimension of 128 x 128 would result in a pitch of 128.
apitrace would copy pitch * height = 128 * 128 byte.

But the allocated buffer might be smaller, as for ATI1 only 128 * 128 / (4 * 4) * 8 bytes would be necessary. For ATI2 128 * 128 / (4 * 4) * 16 bytes are necessary, which matches pitch * height.

Replaying the trace dx9_tex_formats_ati1n.trace crashes on WINE !


The question rises: If Windows advertise a pitch = width, and d3dretrace doesn't crash on dx9_tex_formats_ati1n.trace with current code,
does d3d9 on Windows allocate a buffer of size pitch * height, even if it's not necessary (for ATI1) ?

siro20:
I've done some tests on Windows. You are right, that d3d9 allocates pitch * height bytes for "unknown" formats. I'm going to fix the issue in WINE. This patch is still correct, as we know the required byte count for ATIx. It would reduce the memcpy size for ATI1 by half.

jrfonseca:
Yes, I agree we should limit the memcpy to the minimum.

I tried to prototype this on e21e3b6f830bbdc4dcc2ab3791b3ea0d2520eb3a.  Could you confirm it fixes the problems with WINE/Nine?

jrfonseca:
Fix merged and pushed.

PR #480, opened 2016-08-21
Fix formatting that causes Github to hide words

nicholasbishop:
When github renders markdown it hides words inside of angle brackets. Replacing the angle brackets with backticks fixes the issue.


jrfonseca:
Pushed.  Thanks.

PR #490, opened 2016-09-13
Changing working directory from GUI

### LINKED ISSUE #102: Working directory.
Add option to set target application's working directory.
In Qt, I can done it by set working directory for whole TraceProcess, but it's mean different path for trace-file saving and few more architectural issue.
Anyway, it's good to be able set working directory from CLI tool.

p-podsiadly:
Hello,
this PR adds edit in trace dialog for setting working directory of a traced process. When it is left empty, working directory won't be changed (so it has the same behavior as before). It also adds history of recent launches (File > Recent launches...)
This fixes #102

### DIFF
--- a/gui/mainwindow.cpp
+++ b/gui/mainwindow.cpp
@@ -41,6 +41,7 @@
 #include <QUrl>
 #include <QVBoxLayout>
 #include <QTextBrowser>
+#include <QToolTip>
 
 typedef QLatin1String _;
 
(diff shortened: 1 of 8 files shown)


dor666:
Is there a issue with this PR? I would like to have this functionality.

jrfonseca:
Pushed. Thanks and sorry for the delay.

PR #495, opened 2016-10-21
Tracking the mapped buffer size for retracing

### LINKED ISSUE #494: Crash on retracing an incomplete UBO
Command line:
```
glretrace ../servo/servo.trace --dump-state=1564
```
[servo.trace.zip](https://github.com/apitrace/apitrace/files/545299/servo.trace.zip)

Output:
```
apitrace: warning: caught signal 11
1564: error: caught an unhandled exception
glretrace+0x28ae7c
/lib64/libpthread.so.0+0x10c2f
glretrace+0x99258
glretrace+0x9934c
glretrace+0x9a082
glretrace+0x9cbd2
glretrace+0x55962
glretrace+0x4c622
glretrace+0x4cbc9
glretrace+0x47589
/lib64/libc.so.6: __libc_start_main+0xf0
glretrace: _start+0x28
?
apitrace: info: taking default action for signal 11
Segmentation fault (core dumped)
```

kvark:
Fixes #494

Reason it crashes is because the user provided a buffer object smaller than the UBO structure declared in the shader. This is valid for as long as the shader doesn't access the data outside of the provided region.

### DIFF
--- a/retrace/glstate_shaders.cpp
+++ b/retrace/glstate_shaders.cpp
@@ -419,6 +419,14 @@ dumpUniformBlock(StateWriter &writer,
     if (raw_data) {
         std::string qualifiedName = resolveUniformName(name, size);
 
+        GLuint actual_count = mapping.getSize() / array_stride;
+        if (actual_count < desc.size) {
+            std::cerr
+                << "warning: uniform " << qualifiedName << " has fewer actual elements (" << actual_count
+                << ") than declared as used by the shader (" << desc.size << ")\n";
+            desc.size = actual_count;
+        }
+
         dumpAttribArray(writer, qualifiedName, desc, raw_data + start + offset);
     }
 }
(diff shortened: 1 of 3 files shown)


retrace/glstate_shaders.cpp
```diff
@@ -419,6 +419,14 @@ dumpUniformBlock(StateWriter &writer,
     if (raw_data) {
         std::string qualifiedName = resolveUniformName(name, size);
 
+        GLuint actual_count = mapping.getSize() / array_stride;
```
jrfonseca: Thanks for looking into this.  I agree extra checking is a good thing.

But I fear this check might be a tad too strict, as I'm not sure that the array elements are guaranteed to be contiguous in memory.

Imagine a buffer with 12 bytes, with elements of 4 bytes, and a stride of 8 bytes.  There's space for two elements (with 4 bytes of padding.) But merely diving buffer size by stride would say only one element can fit...

Perhaps this can never happen -- it's hard to say if that's legal or not --, but the safest thing to do would be to pass the buffer size down `dumpAttribArray` and `dumpAttrib`, and do the checking there.

Another alternative would be to try to compensate for this when computing `actual_count`.

PR #497, opened 2016-10-25
glretrace: Show current frame number in window title

aejsmith:
I find this can be useful to roughly identify which frames I want to be
looking at for a specific issue in a large trace, by replaying the trace
and looking at the current frame number when the issue occurs.

### DIFF
--- a/retrace/glretrace_main.cpp
+++ b/retrace/glretrace_main.cpp
@@ -565,6 +565,14 @@ frame_complete(trace::Call &call) {
         retrace::warning(call) << "could not infer drawable size (glViewport never called)\n";
     }
 
+	/* Display the frame number in the window title, except in benchmark mode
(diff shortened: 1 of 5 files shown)


amonakov:
May I suggest extending the patch to suppress title changes when benchmark/profile mode is active, if that's not a bother? Otherwise, window manager and compositor activity will probably affect the profile too much.

Also, I remember noticing frequent window title changes having too much overhead in i3wm. I don't know if that issue is still there, and if it is, it's probably their bug. Not an issue for your patch of course, just something to keep in mind.

aejsmith:
Sure, I've updated the patch to do that.

jrfonseca:
Thanks.  Sorry it took this long.  I keep hoping to go through the backlog of issues/PRs, but never can find the time.

PR #498, opened 2016-11-08
d3d9trace: Capture undocumented ShaderValidator

siro20:
Add support for undocumented IDirect3DShaderValidator9 interface.
Requires updated dxsdk headers.

### DIFF
--- a/specs/d3d9.py
+++ b/specs/d3d9.py
@@ -449,3 +449,18 @@ d3dperf.addFunctions([
     StdFunction(Void, "D3DPERF_SetOptions", [(DWORD, "dwOptions")], sideeffects=False),
     StdFunction(DWORD, "D3DPERF_GetStatus", [], fail='0', sideeffects=False),
 ])
+
+IDirect3DShaderValidator9 = Interface("IDirect3DShaderValidator9", IUnknown)
+PDIRECT3DSHADERVALIDATOR9 = ObjPointer(IDirect3DShaderValidator9)
+
+# undocumented interface
+IDirect3DShaderValidator9.methods += [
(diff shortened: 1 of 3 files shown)


specs/d3d9.py
```diff
@@ -449,3 +449,18 @@
...
+# undocumented interface
+IDirect3DShaderValidator9.methods += [
+    StdMethod(HRESULT, "Begin", [(OpaquePointer(Void), "arg1"), (OpaquePointer(Void), "arg2"), (ULONG, "arg3")]),
+    StdMethod(HRESULT, "Instruction", [(OpaquePointer(Void), "arg1"), (UINT, "arg2"), (ULONG, "arg3"), (UINT, "arg4")]),
+    StdMethod(HRESULT, "End", []),
+]
+
+d3d9shadervalidator = Module("d3d9")
+d3d9shadervalidator.addFunctions([
+    StdFunction(PDIRECT3DSHADERVALIDATOR9, "Direct3DShaderValidatorCreate9", [], fail='NULL'),
```
jrfonseca: If you don't really care about the `IDirect3DShaderValidator9` interface, and just want to fullfil the `Direct3DShaderValidatorCreate9` entrypoint, you can simply do:

    StdFunction(PVOID, "Direct3DShaderValidatorCreate9", [], internal=True),

and remove everything else.

jrfonseca:
@siro20, as explained in https://github.com/apitrace/dxsdk/pull/1#issuecomment-259279977 this is not the right way of going about this.  If you provide more context of why you need to trace `Direct3DShaderValidatorCreate9` I might be to set you on the right direction.

PR #512, opened 2017-03-21
retrace: fixed DDLOCK_DONOTWAIT flag clear mask

simonowen:
This was unintentionally clearing all other lock flags, as well as
potentially having both wait and do-not-wait set simultaneously.


jrfonseca:
Good find.  I screwed up when I tried to clean up this up on c28e606f7656de3a3a8808b68eb048528397d5a7.

PR #516, opened 2017-04-10
Fixed issue with state dumping and multithreading

MartinSherburn:
Fixed issue where state could be dumped before the requested call had been executed in a multithreaded trace (for example when getting state for a particular Draw call, the framebuffer could contain the contents prior to the Draw call)

### DIFF
--- a/retrace/retrace_main.cpp
+++ b/retrace/retrace_main.cpp
@@ -280,7 +280,8 @@ retraceCall(trace::Call *call) {
         }
     }
 
-    if (call->no >= dumpStateCallNo &&
+    // dumpStateCallNo is 0 when fetching default state
+    if ((call->no == dumpStateCallNo || dumpStateCallNo == 0) &&
         dumper->canDump()) {
         StateWriter *writer = stateWriterFactory(std::cout);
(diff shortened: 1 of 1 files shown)


MartinSherburn:
It appears that with multi-threaded traces the calls don't get replayed sequentially. So if a later call gets replayed before the targeted call, it would dump the state prematurely and exit the program. I noticed this when the framebuffer for the draw calls I was trying to dump state for did not contain the result of the draw operation.

MartinSherburn:
What about:

```
if ((call->no == dumpStateCallNo || dumpStateCallNo == 0) &&
          dumper->canDump()) {
...
```
That would solve both issues, if you think it's a good idea I'll update the pull request.

jrfonseca:
I've finally merged this. 
I did a subsequent tweak so that glretrace exits early with an error if the requested call was found, but it was impossible to dump the state (b9e4b73ced4e9547bb1edd1d4f452fffcd350630) otherwise glretrace would continue replaying all the end till the end unnecessarily.
But otherwise I agree this is a sensible thing to do.
Thanks and sorry for the delay.

PR #540, opened 2017-11-26
retrace: image/texture handles may be in different spaces

imirkin:
Bindless handles aren't portable across implementations. They may be
passed in via vertex attributes, uniforms, etc. We still keep the maps
around to limp along for making handles resident.


jrfonseca:
Thanks.

Yep, it's indeed hard to deal with bindless handles effectively.

I'm CC'ing @ahuillet who was also interested in this at one point.

One way or the other apitrace will need cooperation from the OpenGL implementation to be able to replay traces with bindless handles:
- make handles reproducible somehow
- allow apitrace to tell implementation to which handles it should use
- make handles in someway _unique_ or rare (e.g, XOR in with some unusual pattern) so we could search'n'replace mappings for the handles 

Not sure what else..

imirkin:
The fact that they can be passed in via a vertex buffer or ssbo or ubo or anything else makes it pretty much impossible to track anything down in a generic manner.

Nothing prevents an application from taking an array of handles, gzipping them, feeding that into a ssbo, and then decode the gzipped data in a shader and use the handles that way via the uvec2 <-> sampler/image casting. I'm not suggesting that this is common, or a good idea, but just taking my point to the extreme.

It's actually the application which has the knowledge of how all this stuff is used, so ideally there'd be some way for the application to make explicit remap calls. But that won't happen... not sure what can be done in the apitrace / driver stack combination.

imirkin:
The only thing which could work *reliably* was if the driver were aware of the remapping tables, and in the *shader* performed the remapping lookup. Of course shaders aren't a great place to implement this sort of thing, it'll obviously be a bunch of inefficient code. But it could work - there aren't actually a lot of handles (not like millions), so the size of the remapping table wouldn't be that big. Some simple trick to do a hash + walk type of approach should be doable, although it'll tank perf. (But then again, so do texture lookups, so nothing's free in life.)

It would mean that apitrace would have to inform the driver via some additional ext about the remapping, and it would be on the driver to implement the remapping logic in all the shaders whenever bindless handles are used.

This is the only reliable mechanism I can think of to enable bindless to work across all implementations.

PR #542, opened 2017-12-03
dxgiretrace: Allow NULL pointers for WKPDID_D3DDebugObjectName.

jozefkucia:
Some apps set WKPDID_D3DDebugObjectName private data to NULL.


jrfonseca:
Thanks.

PR #544, opened 2017-12-06
MSAA Texture Resolve / Unresolve surface feature with GUI checkbox

gabedagani:
Squashed commits into 1. Fixed early returns.

Provide msaa texture resolve feature with GUI checkbox. Requires a retrace to change the output and uses the CLI argument --no-msaa-resolve to fetch all the sample frames in a vertical array. Shows up in the surfaces tab of the texture.


gabedagani:
closing, will rebranch single commit.

PR #545, opened 2017-12-06
MSAA Texture Resolve / Unresolve surface feature with GUI checkbox

gabedagani:
Add feature in single commit. rebased.


jrfonseca:
Pushed. Thanks.

PR #551, opened 2018-01-09
Fix some warnings found when using MSAA for depth/Stencil

gabedagani:
Avoid looking up state on unbound features. Make sure to replace state for GL_READ and GL_DRAW buffers.


jrfonseca:
Pushed.  Thanks.

PR #555, opened 2018-02-13
glretrace: Implement --borderless option for xlib

davidlunarg:
Here is a pull request that implements the --borderless option. We have found it useful for replaying traces that were created from an app that runs in a full-screen borderless window.

Unfortunately, this is implemented for xlib only. I don't think it would be that difficult to extend this for other platforms.


jrfonseca:
It looks like macOS build is still broken.

jrfonseca:
BTW, forgot to mention, I agree this is a nice idea.  In fact I felt the need for something similar for windows.

davidlunarg:
Yep, when I submitted this pull request, I found that MacOS didn't compile. I didn't have a MacOS system to try out fixes on, so I put it aside. Presently, this is a lower priorty for me, I'll get to it sometime in the near future.

Issue #436, opened 2016-03-21
Support reading float buffers and depth buffers for d3d 9


axeldavy:
It would be great if we could see the content of float types textures and render targets (R16F, R32F).

Also it would be great to be able to see the depth buffer content. However it is likely more complicated: D32F_lockable seems widely supported, but there doesn't seem to be easy way to copy the depth buffer content to it.

EDIT:
Oh, I see you are using DXGI for a wide range of format support. Would you accept a pull request to support more formats without DXGI ?

jrfonseca:
> Oh, I see you are using DXGI for a wide range of format support.

Right, through DirectXTex library, which is DXGI/D3D10+.

> Would you accept a pull request to support more formats without DXGI?

Probably adding hand-written code paths for some formats is unavoidable.  But I'd like to avoid reinventing the wheel as much as possible.

Could you propose a change that uses D3DX9's D3DXLoadSurfaceFromSurface to do the conversion when DXGI is not available?  I don't know about WINE's, but the real D3DX9 library supports lots of format conversion.

jrfonseca:
> Also it would be great to be able to see the depth buffer content. However it is likely more complicated: D32F_lockable seems widely supported, but there doesn't seem to be easy way to copy the depth buffer content to it.

INTZ has the same bit pattern as D24S8, is widely supported, http://aras-p.info/texts/D3D9GPUHacks.html , and it can be locked (I just tried it on NVIDIA.)

However, the semantics of sampling from INTZ and D24S8 are different -- the former samples raw z values, where as the latter does implicit shadow compre.  And there's no practical way to blit from D24.

Microsoft PIX managed to get D3D9 depth values some how.  I don't know if it used some sort of backdoor to bypass the D3D9 runtime limations.

The only effective way to read an arbitrary depth-stencil buffer without abusing the D3D9 API is to:
* Create two `D3DFMT_R32F` surfaces with the same size as the depth stencil buffer, bind one as rendertarget, the other as sampler.
* For every bit in the depth buffer draw a textured quad, _bisecting_ the depth value using the depth test 
* First draw a quad with Red = Z = 0.5, ZFUNC=GREATEREQUAL.  This gets if MSB of the Z buffer
* Blit from render-target to texture
* Draw a quad with Red = 0.25, Z = 0.25 + sample(), and BLEND=ADD . This get the second MSB
* and so on.

A similar technique can be used for recovering the stencil values.

[...]

PR #558, opened 2018-02-25
D3D9 depth dump

siro20:
Fix for #436.
Dump depth buffers using RESZ hack.
Tested on Windows and WINE.

### DIFF
--- a/retrace/d3d9state_images.cpp
+++ b/retrace/d3d9state_images.cpp
@@ -35,6 +35,8 @@
 #include "d3dstate.hpp"
 
 
+#include <vector>
+
 namespace d3dstate {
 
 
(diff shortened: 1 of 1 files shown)


jrfonseca:
This didn't work for me when I tried on Windows on Intel HD 4000 (wrong depth values.)  And I couldn't get it to work on WINE neither (segfaults, though unrelated.)

Still, it definetely seems better than nothing, and perhaps the outstanding issues can be ironed out at some stage.

So it's now pushed.  Thanks.

jrfonseca:
One final comment regarding implementation: use VS might yield more reliable beheavior, as most state doesn't matter in that case.

PR #559, opened 2018-02-26
minor fixes found on QNX

groleo:


groleo:
right.
the specs says RTLD_ should be macros:
http://pubs.opengroup.org/onlinepubs/009695399/basedefs/dlfcn.h.html
I'll update the PR.
thx

groleo:
Opened a new PR:
https://github.com/apitrace/apitrace/pull/571

PR #562, opened 2018-04-13
Add call filtering for sed command

sercxanto:
"trace sed" now accepts a parameter like the trim command:

    --calls=CALLSET parameter

It applies the sed renaming only to those callsets specified.

Use case: Rename enums only for certain specific calls, not globally in
the whole trace.


jrfonseca:
Thanks.

PR #566, opened 2018-05-09
wrapper: Also generate DLL interposers for D3D11 tracing.

Plagman:
This lets me trace D3D11 games that don't want to run through apitrace.exe such as Far Cry 5 and Assassin's Creed: Origins.


Plagman:
My new files didn't tag along for the ride somehow, will fix..

jrfonseca:
I've worked around the MSBuild woes and pushed this.

I think this way of injecting will come quite handy.  Thanks!

PR #567, opened 2018-05-23
glapi: Add support for glReadnPixelsKHR from KHR_robustness

gerddie:
This adds another name for glReadnPixels that might be used (e.g. virglrenderer uses it). 

Thanks for considering the patch.


jrfonseca:
Thanks.

PR #568, opened 2018-05-23
retrace: new parameter min-cpu-time

groleo:


jrfonseca:
Sorry for the long delay.  I really haven't had time to maintain apitrace.

Maybe I'm missing something, but I don't understand what this new parameter does from the change.  You pass around the command line parameter value, but what is it used for?

groleo:
Hi Jose,

minCpuTime is used to filter calls which takes at least minCpuTime; the
thing is that it's always set to 1000, so
the commit is adding a way of setting the minCpuTime from command line, for
the case where one is interested
to see calls that take less than 1000.

Let me know if it makes sense to update the commit message.

thanks

jrfonseca:
Make sense now.  The only thing missing was updating the `usage()` function, but I just went ahead and did that.

Thanks

PR #604, opened 2019-02-06
Add envvar FLUSH_EVERY_MS.

kdashg:
If present, FLUSH_EVERY_MS forces apitrace to flush trace files at
regular intervals.


jdashg:
In particular, when tracing (multiprocess) Firefox, background process traces don't seem to flush to disk properly.

jrfonseca:
I agree this is a sensible thing to do. In fact, it probably makes sense to do it by default.

But this needs to be done in a way it doesn't at least break other OSes build.  And I'm afraid C++11 threading primitives are not quite to a point we can use them directly.  Please see the threading abstractions in `lib/os/os_thread.hpp`.

jdashg:
Seems to compile fine after removing those includes to rely on os_thread.hpp.
PTAL!

chealy:
I tested this change as I was facing similar issues with my env.  Change works as described.  It would be good to see this code merged.

Tested-by: Chris Healy <cphealy@gmail.com>

PR #605, opened 2019-02-20
Add an option to use regex when searching in qapitrace

10110111:
This commit adds a checkbox in the search widget to use regular expressions when searching. This lets in particular find exact numbers: e.g. `\b204\b` would match the number `204` but not `1.7204e-3`. Another use case: search for calls using patterns like `gl(Named)?BufferData`.


jrfonseca:
Pushed. Thanks!

PR #610, opened 2019-04-29
Unbundle brotli

a17r:


jrfonseca:
I'm okay using system brotli when available.  But I'd like to retain the ability of using the bundled brotli if no system brotlie is available.  So this PR needs to be updated to do so.

See the existing CMake logic for zlib and libpng as reference.

jrfonseca:
3b4a698dde40022628a2e944008c43a67a6e1c02 addressed this issue.

PR #611, opened 2019-04-30
Fix out-of-source build without bundled snappy

iDarkTemplar:


iDarkTemplar:
Related: https://bugs.gentoo.org/684802

This minor change fixes build issue, but the build failure also reveals another issue: header file names collision.

Both bundled snappy and libbacktrace have file named config.h. Currently, wrong header might be included during build for either snappy or libbacktrace, it might be a good idea to rename those files to get rid of collision or make sure both directories containing those files are not included at once if that's possible and renaming headers is undesirable.

jrfonseca:
Thanks.  Pushed and fixed snappy too.

It's unfortunate that libbacktrace depends on config.h from the public headers.  But thankfully snappy only depends on config.h from internal source files.

PR #615, opened 2019-05-23
dxgitrace: Fix a thread safety of a d3d11 CreateBuffer method

asimiklit:
According to directx 11 spec all device methods are thread safe:
   "All ID3D11Device interface methods are free-threaded, which means
    it is safe to have multiple threads call the functions
    at the same time."

But all device-context methods according to directx 11 spec:
   "Any context – immediate or deferred – can be used on any thread
    as long as the context is only used in one thread at a time."

So without this fix WrapID3D11Device*::CreateBuffer methods are not
thread safe. It could lead to the crash in some games which are
create buffers in different threads.

Signed-off-by: Andrii Simiklit <andrii.simiklit@globallogic.com>

### DIFF
--- a/wrappers/dxgitrace.hpp
+++ b/wrappers/dxgitrace.hpp
@@ -79,51 +79,37 @@ _shouldShadowMap(ID3D11Resource *pResource)
  */
 
 inline void
-_initializeBuffer(ID3D10Device *pDevice, const D3D10_BUFFER_DESC *pDesc, ID3D10Buffer *pBuffer)
+_initialBufferAlloc(const D3D10_BUFFER_DESC *pDesc, D3D10_SUBRESOURCE_DATA *pInitialData)
 {
-    HRESULT hr;
+    void *data = malloc(pDesc->ByteWidth);
+    assert(data);
+    MemoryShadow::zero(data, pDesc->ByteWidth);
 
-    if (!(pDesc->CPUAccessFlags & D3D10_CPU_ACCESS_WRITE)) {
-        return;
-    }
-
-    assert(pBuffer);
-
-    D3D10_MAP MapType = pDesc->Usage == D3D10_USAGE_DYNAMIC ? D3D10_MAP_WRITE_DISCARD : D3D10_MAP_WRITE;
-    VOID *pData;
-    hr = pBuffer->Map(MapType, 0, &pData);
-    assert(SUCCEEDED(hr));
-    if (FAILED(hr)) {
-        return;
-    }
-
-    MemoryShadow::zero(pData, pDesc->ByteWidth);
-    pBuffer->Unmap();
+    pInitialData->SysMemPitch = 0;
+    pInitialData->SysMemSlicePitch = 0;
+    pInitialData->pSysMem = data;
 }
 
 inline void
-_initializeBuffer(ID3D11Device *pDevice, const D3D11_BUFFER_DESC *pDesc, ID3D11Buffer *pBuffer)
+_initialBufferFree(D3D10_SUBRESOURCE_DATA *pInitialData)
 {
-    HRESULT hr;
+    free(const_cast<void*>(pInitialData->pSysMem));
+}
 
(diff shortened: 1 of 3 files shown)


asimiklit:
I faced with some crashes while I was trying to trace a "NieR: Automata" and "Skyrim Special Edition" (it happened in a gameplay phase). I tested it on ubuntu 19.10 using wine + cross compiled apitrace wrappers.

I found out that it is a multithreading issue and also I found on MSDN that d3d11 device methods must be thread-safe so we can't to use thread-unsafe device context methods in device methods but apitrace do it. I tried just to remove the _initializeBuffer call from CreateBuffer method in d3d11 device wrapper and it helped to avoid crashes.

jrfonseca:
Great find!

Ironically just a couple of days ago I found another different bug with precisely this code.

Currently we only initialize buffers that are CPU accessible (ie, can be mapped.)  But itturns out that some applications (AutoCAD 2018 in particular) create a non-mappable buffer, and copies _unitialized_ data from that first buffer into a staging buffer, tricking Apitrace to think nothing was written, when the writing contents matches precisely the uninitialised contents.  This causes the trace to lack important memcpy calls, which make it useless.

That means apitrace should initialize *all* buffers, and not just the CPU accessible buffers. Consequently, the best way of doing that is to modify `dxgitrace.py` to always set `pInitialData` argument instead.

That is, when `CreateBuffer` methods are called, `pInitialData` is null, we should just allocate some memory equal to the buffer size, fill it with the special pattern, and pass it.

This would also fix the issue you reported, as we wouldn't need to resort to non-thread-safe `ID3D11DeviceContext` methods.

What do think?

asimiklit:
Thanks a lot for the advice, you are right it is much better solution.
I will provide reworked solution shortly.

One more question:
I found that almost all d3d11 device methods call: Wrap D3D11****::_wrap(..)
All these methods use g_WrappedObjects without any locks.
It also is thread-unsafe behavior. 
What do you think about such solution for it?
```
diff --git a/wrappers/trace.py b/wrappers/trace.py
index 5618e4b4..e0f621e6 100644
--- a/wrappers/trace.py
+++ b/wrappers/trace.py
@@ -529,6 +529,7 @@ class Tracer:
         print()
         print()
         print('static std::map<void *, void *> g_WrappedObjects;')
+        print('static os::mutex g_WrappedObjectsLock;')
 
     def footer(self, api):
         pass
@@ -816,6 +817,8 @@ class Tracer:
         print(r'        return;')
         print(r'    }')
         print(r'    assert(hasChildInterface(IID_%s, pObj));' % iface.name)
+        if wrapperInterfaceName.upper().find("D3D11") >= 0:
+            print(r'    os::unique_lock<os::mutex> l(g_WrappedObjectsLock);')
         print(r'    std::map<void *, void *>::const_iterator it = g_WrappedObjects.find(pObj);')
         print(r'    if (it != g_WrappedObjects.end()) {')
         print(r'        Wrap%s *pWrapper = (Wrap%s *)it->second;' % (iface.name, iface.name))
```

jrfonseca:
> I will provide reworked solution shortly.

Thanks!

> All these methods use g_WrappedObjects without any locks.

There is a mutex in `trace::LocalWriter::mutex`, that should be held already when accessing `g_WrappedObjects`.  This lock is hold during the `beginEnter/endEnter` and `beginLeave/endLeave` sections, ie, which serializing the arguments.  So it shouldn't be necessary, if I'm not mistaken.

This lock is released when calling the function, which is why this issue arose in the first place (ie, the CreateBuffer and Map were being called outside the lock.).

Even if another lock isn't needed, I admit there should be at least a comment about this before `g_WrappedObjects`.

asimiklit:
Thanks for clarification) Will add comment.

asimiklit:
It is done. Maybe you have some suggestions or it is ok for you?

jrfonseca:
`interface.name.upper().find("D3D11") >= 0` could be replaced with `interface.name.startswith("ID3D11")`

Otherwise looks great.  I've tweaked and pushed.  Thanks!

asimiklit:
Thanks :)

PR #616, opened 2019-05-29
memtrace: Update nBlocks when pointer changes in MemoryShadow::cover

andrew-krv:
During each ::Map different pointers could be returned by DirectX.
Thus for the same resource with the same subresource having the same size
pointer may change between calls of MemoryShadow::cover. With the pointer
change number of blocks also may change, which may later cause memory
access violation in MemoryShadow::update.

### DIFF
--- a/wrappers/memtrace.cpp
+++ b/wrappers/memtrace.cpp
@@ -184,7 +184,7 @@ void MemoryShadow::cover(void *_ptr, size_t _size, bool _discard)
 {
     assert(_ptr);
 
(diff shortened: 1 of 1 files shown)


andrew-krv:
This was made to fix Skyrim Special edition crash, while tracing using Windows.

jrfonseca:
Great find!

I wonder if this explains some issues that existed before with deferred contexts mappings, as I suspect that in that case the runtime will return a new pointer on every discard.

PR #618, opened 2019-06-06
dxgitrace: Fix null pointer dereference in createWindowForSwapChain

werman:
CreateDeviceAndSwapChain family of functions can accept null pSwapChainDesc.

Found when tried to record and replay a trace of looped RenderDoc's trace to find what call hangs the driver.  Call to CreateDeviceAndSwapChain in RenderDoc - [renderdoc/driver/d3d11/d3d11_replay.cpp#L3701](https://github.com/baldurk/renderdoc/blob/e4333291a112f9fbf5491e367ae900ddc837ee20/renderdoc/driver/d3d11/d3d11_replay.cpp#L3701)


jrfonseca:
Thanks.

PR #620, opened 2019-06-20
Use Python3_EXECUTABLE in cmake.

jdashg:


jdashg:
This fixed my build issues on MacOS.
Previously, I was getting "permission denied: glproc.py" and other errors.

jrfonseca:
Hmm it seems to cause problems with older CMake (see travis).  Do you know CMake version this Python3 module works reliably?

jdashg:
Looks like CMake 3.12.

jrfonseca:
Thanks. I pushed an updated version of your change, which keeps the old behavior for older CMake.

Hopefully this gets the best of both worlds.

PR #623, opened 2019-07-22
Support writable coherent buffers

### LINKED ISSUE #232: apitrace: coherent mapping not implemented
Both ARB_buffer_storage and AMD_pinned_memory allow to use coherent memory in shaders. As apitrace doesn't catch writes in this memory, the results are likely wrong.

AMD_pinned_memory could easily be disabled, but as ARB_buffer_storage is required for GL4.4, it's needed to be emulated.

tbh, I don't see any os Independent way to implement this but to have a second buffer synced before/after almost every call.
But there are useful os dependent ways: eg we could alloc a both write and read protected memory and only sync the accessed pages.

werman:
This implements handling of writeonly and read-write coherent buffers. Detailed descriptions could be found in respective commit messages.

Would close #232 

### Current shortcomings:
- Buffer's contents may be changed from inside the shaders however this wouldn't be reflected in shadow memory and if the same page is changed on CPU side and on GPU side even if offsets are different the data on the GPU will be overwritten.

- Small changes in buffer result in full page write.

- All readable buffers are synced back to cpu on fence signal instead of only potentially changed ones. I did not found anything that writes and reads from the same buffers with the exception of Wine which would do this if DX11 game would to such thing. Since the case is rare I didn't want to add complexity. 

### What is tested
On Linux:
- CS:GO - works
- Bioshock Infinite - works
- Super Tux Kart - works
- Some synthetic tests from piglit and several other random tests - work
- Dota2 - doesn't, it shows several good frames and after that some vertices start having wrong values resulting in many flashing polygons. And I'm not able to identify the cause.
- DiRT Showdown - didn't test
- Wine with some DX11 app using coherent buffers - didn't test 

On Windows I've tested with Super Tux Kart to check the platform specific path and tracing works.

### Tests

Tests for apitrace-tests  would hopefully come soon.

### Some code considerations:

Application may write from several threads to coherent buffers so I've added a broad mutex in relevant places and since I didn't see anyone heavily writing from several threads simultaneously I didn't do any optimizations for it. The only game that wrote from several threads was Dota2.

### DIFF
--- a/wrappers/glmemshadow.cpp
+++ b/wrappers/glmemshadow.cpp
@@ -343,22 +343,26 @@ void GLMemoryShadow::commitWrites(Callback callback)
         }
     }
 
-    if (isPageDirty(mappedStartPage)) {
-        const size_t shadowOffset = mappedStart % sPageSize;
-        const size_t size = std::min(sPageSize - glStartOffset, mappedSize);
-
-        memcpy(glMemory, shadowSlice + shadowOffset, size);
-        callback(shadowSlice + shadowOffset, size);
-    }
-
-    for (size_t i = mappedStartPage + 1; i < mappedEndPage; i++) {
+    for (size_t i = mappedStartPage; i < mappedEndPage; i++) {
         if (isPageDirty(i)) {
-            const size_t shadowOffset = (i - mappedStartPage) * sPageSize;
-            const size_t glOffset = shadowOffset - glStartOffset;
-            const size_t size = std::min(glStartOffset + mappedSize - shadowOffset, sPageSize);
+            // We coalesce consecutive writes into one
+            size_t firstDirty = i;
+            while (++i < mappedEndPage && isPageDirty(i)) { }
 
-            memcpy(glMemory + glOffset, shadowSlice + shadowOffset, size);
-            callback(shadowSlice + shadowOffset, size);
+            const size_t pages = i - firstDirty;
+            if (firstDirty != mappedStartPage) {
+                const size_t shadowOffset = (firstDirty - mappedStartPage) * sPageSize;
+                const size_t glOffset = shadowOffset - glStartOffset;
+                const size_t size = std::min(glStartOffset + mappedSize - shadowOffset, sPageSize * pages);
+
+                memcpy(glMemory + glOffset, shadowSlice + shadowOffset, size);
+                callback(shadowSlice + shadowOffset, size);
+            } else {
+                const size_t size = std::min(sPageSize * pages - glStartOffset, mappedSize);
+
+                memcpy(glMemory, shadowSlice + glStartOffset, size);
+                callback(shadowSlice + glStartOffset, size);
+            }
         }
     }
 
--- a/wrappers/glmemshadow.cpp
+++ b/wrappers/glmemshadow.cpp
@@ -1,0 +1,408 @@
+/*
+ * Copyright © 2019 Intel Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


wrappers/glmemshadow.cpp
```diff
@@ -0,0 +1,349 @@
...
+    uint8_t *shadowSlice = shadowMemory + mappedStartPage * sPageSize;
+    const size_t glStartOffset = mappedStart % sPageSize;
+
+    if (isPageDirty(mappedStartPage)) {
+        const size_t shadowOffset = mappedStart % sPageSize;
+        const size_t size = std::min(sPageSize - glStartOffset, mappedSize);
+
+        memcpy(glMemory, shadowSlice + shadowOffset, size);
+        callback(shadowSlice + shadowOffset, size);
+    }
```
sergiiromantsov: I have a feeling that here should be:
`shadowSlice += shadowOffset`

wrappers/gltrace.py
```diff
@@ -514,6 +593,7 @@ def traceFunctionImplBody(self, function):
             print('                flush = flush && flushing_unmap;')
             print('            }')
             print('            if (flush && length > 0) {')
+            print('            ')
```
sergiiromantsov: seems redundant?

werman: indeed

wrappers/gltrace.py
```diff
@@ -688,6 +797,7 @@ def traceFunctionImplBody(self, function):
 
         Tracer.traceFunctionImplBody(self, function)
 
+
```
sergiiromantsov: one more redundant

wrappers/glmemshadow.cpp
werman: For the first dirty page we may need to start copying from `shadowSlice` with an offset depending on the mapping start offset and will copy `pageSize - offset` bytes.

For the next page the copying will start from the start of the page in shadowSlice which is `shadowSlice + pageSize` and would copy into `glOffset + pageSize - glStartOffset`.

So no, there is no need to change `shadowSlice` pointer.

sergiiromantsov: yes, you are right

wrappers/glmemshadow.cpp
```diff
@@ -0,0 +1,404 @@
...
+}
+
+void memProtect(void *addr, size_t size, MemProtection protection) {
+#ifdef _WIN32
+    DWORD flOldProtect;
+    BOOL bRet = VirtualProtect(addr, size, static_cast<DWORD>(protection), &flOldProtect);
+    if (!bRet) {
+        DWORD dwLastError = GetLastError();
+        os::log("apitrace: error: VirtualProtect failed with error 0x%lx\n", dwLastError);
+        exit(EXIT_FAILURE);
```
jrfonseca: `exit()` might still call C++ destructurors.  It's better on wrapper code to always call `os::abort()`

wrappers/glmemshadow.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    size_t nPages = 0;
+
+    size_t mappedStartPage = 0;
+    size_t mappedEndPage = 0;
+
+    bool isDirty = false;
+    std::vector<uint32_t> dirtyPages;
+    uint32_t pagesToDirtyOnConsecutiveWrites = 1;
+    uint32_t lastDirtiedRelativePage = ~0 - 1;
```
jrfonseca: `~0` -> `UINT32_MAX`, just for consistency.

wrappers/gltrace.py
```diff
@@ -646,6 +741,13 @@ def traceFunctionImplBody(self, function):
             print(r'        os::log("apitrace: warning: GL_INTEL_map_texture not fully supported\n");')
             print(r'    }')
 
+
```
jrfonseca: Only one empty line. Also a comment would be nice.

wrappers/glmemshadow.cpp
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    for (size_t i = mappedStartPage + 1; i < mappedEndPage; i++) {
+        if (isPageDirty(i)) {
+            const size_t shadowOffset = (i - mappedStartPage) * sPageSize;
+            const size_t glOffset = shadowOffset - glStartOffset;
+            const size_t size = std::min(glStartOffset + mappedSize - shadowOffset, sPageSize);
+
+            memcpy(glMemory + glOffset, shadowSlice + shadowOffset, size);
+            callback(shadowSlice + shadowOffset, size);
```
jrfonseca: Rather that emitting one fake mempcy to the trace per page, it would be more efficient to coalesce all writes to consecutive pages into a single fake memcpy.

wrappers/glmemshadow.cpp
werman: Didn't think about this, thanks.

werman:
> Please confirm: there's no behavior change for non-coherent mappings right?

Yes, there is no behavior change for non-coherent or readonly coherent mappings. I've checked everywhere for buffer to have `GL_MAP_COHERENT_BIT | GL_MAP_WRITE_BIT`.

werman:
I've made the write coalescing as a separate commit.

Dota2 is now could be correctly traced because I moved protection of the dirty pages against writes _before_ reading from them instead of after, thus other threads cannot override data in buffer while we copy it.

jrfonseca:
Merged.  Thanks!

PR #625, opened 2019-08-13
fix crash in d3d shaderSize function

megai2:
Removed comment skipping due to false detection of comment opcode.


jrfonseca:
> Removed comment skipping due to false detection of comment opcode.

That's odd.  How could we be erroneously detecting a comment opcode?  Could you please provide me the trace containing the shader that triggered this?

jrfonseca:
Nevermind, I see the problem.  The issue is the

    DWORD dwToken = pFunction[dwLength++];

line.

Instead of blindly incrementing dwLength one by one, we should be taking in consideration the bits pointed by `D3DSI_INSTLENGTH_MASK`  and `D3DSI_INSTLENGTH_SHIFT`.

Your fix is still not 100% correct, as we could have the value of `D3DSIO_END` in the middle of the body too.

megai2:
With this fix trace will not crash at least.

100% correct one should not only read instruction length, but also lookup that instruction length in case of vs/ps model used is lower than 2_0.

I can put table with instruction sizes for that case, but i'm not sure about 1_x opcodes length.
Are they 100% constant in size or have some difference in size with relation of opcode?

jrfonseca:
Thanks for the revised patch.

I took a closer look into this, as I fear getting this wrong might cause extensive breakage.

For SM 2.0 shaders we can indeed use `D3DSI_INSTLENGTH_MASK/SHIFT`, and that's the most reliable.

But it also seems that for SM 1.x shaders we don't need complicated parsing -- we can just use bit 31 to detect the beginning of instruction tokens.

Try this patch: [pull625.txt](https://github.com/apitrace/apitrace/files/3527120/pull625.txt)

megai2:
Seems to be fair enough, but i still don't have sm 1_x shaders to trace upon.

jrfonseca:
You make good points in your last comment.  DEFB/DEFI are SM 2.x, so we don't need to worry about them, but DEF is indeed another problem.   I'll incorporate your comment/change and merge.

jrfonseca:
Merged.  Thanks!

PR #626, opened 2019-08-19
gltrace: compressed textures size

sergiiromantsov:
During replay of Dota 2 observed crash.
It was happening due invalid memory access.
Observed that for compressed textures a real
memory buffer is computed from
GL_UNPACK_ROW_LENGTH but not from imageSize.

Checked only with glCompressedTexSubImage2D, but others
should have similar logic.

Signed-off-by: Sergii Romantsov <sergii.romantsov@globallogic.com>


jrfonseca:
@sergiiromantsov I'm not sure I understand.  Shouldn't apps list the true dta size in the `imageSize` parameter?

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glCompressedTexSubImage2D.xhtml clearly states that `imageSize` "Specifies the number of unsigned bytes of image data starting at the address specified by data."

Are you saying that Dota 2 is passing a bogus / too small value on imageSize?

sergiiromantsov:
> https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glCompressedTexSubImage2D.xhtml clearly states that `imageSize` "Specifies the number of unsigned bytes of image data starting at the address specified by data."

Yes, but OpenGL p. 8.7. COMPRESSED TEXTURE IMAGES also says:
"If the compressed data are arranged into fixed-size blocks of texels, the pixel
storage modes can be used to select a sub-rectangle from a larger containing rect-
angle. These pixel storage modes operate in the same way as they do for Tex-
Image*D and as described in section 8.4.4.
....
There is a mechanism for selecting a sub-rectangle of groups from a
larger containing rectangle. This mechanism relies on three integer parameters:
UNPACK_ROW_LENGTH , UNPACK_SKIP_ROWS , and UNPACK_SKIP_PIXELS . Be-
fore obtaining the first group from memory, the data pointer is advanced by
( UNPACK_SKIP_PIXELS )n + ( UNPACK_SKIP_ROWS )k elements. Then width
groups are obtained from contiguous elements in memory (without advancing the
pointer), after which the pointer is advanced by k elements. height sets of width
groups of values are obtained this way. See figure 8.2."

> Are you saying that Dota 2 is passing a bogus / too small value on imageSize?
Seems no, OpenGL p. 8.7. COMPRESSED TEXTURE IMAGES says:
"For CompressedTexImage2D the imageSize parameter must be equal
to blocksize × (width / bw) * (height / bh)".
In our case for format GL_COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT its: 16 * (256 / 4) * (256 / 4) == 65536 == imageSize


Maybe you could suggest some games which uses Compressed(Sub)TexImage2/3D? - I whould check them also.

janesma:
Civilization VI crashes during trace with this patch.

werman:
> Civilization VI crashes during trace with this patch.

Interesting, I'll check next week what's wrong. Hope it's something trivial, with Dota 2 being the only real test case it was hard to be sure in correctness.

werman:
I fixed crash in Civilization VI and add handling of `GL_UNPACK_SKIP_IMAGES` however with the lack of tests I'm not entirely sure if handling of depth is correct.

jrfonseca:
I'm still not convinced this is right...

AFAICT, the `imageSize` is the amount of bytes pointed by `data`.  Period.  It doesn't matter if application uses `GL_UNPACK`, it should pass the final size in `imageSize`.

IIUC, it seems to me that this is an application bug.  Civilization is passing `imageSize` with the full image size, but it's selecting a smaller rectangle via UNPACK, and worse, passing less bytes that `imageSize` in `data`.

And what's done here could easily create bugs on correct applications, especailly if the imageSize value we compute ends up being wrong due to some funny GL extension.

werman:
> I'm still not convinced this is right...
> 
> AFAICT, the `imageSize` is the amount of bytes pointed by `data`. Period. It doesn't matter if application uses `GL_UNPACK`, it should pass the final size in `imageSize`.
> 

The spec clearly states that `imageSize` depends only on width, height, depth and block properties (OpenGL 4.6 "8.7. COMPRESSED TEXTURE IMAGES" p. 228):

image
 
1D, 2D are similar. 

That's indeed very confusing.

> IIUC, it seems to me that this is an application bug. Civilization is passing `imageSize` with the full image size, but it's selecting a smaller rectangle via UNPACK, and worse, passing less bytes that `imageSize` in `data`.
> 

Small correction Civilization is the one which passes size which is enough for us (since it doesn't pass compression parameters this patch initially didn't work with it), it's Dota 2 which uses UNPACK params.

> And what's done here could easily create bugs on correct applications, especailly if the imageSize value we compute ends up being wrong due to some funny GL extension.

jrfonseca:
> The spec clearly states that imageSize depends only on width, height, depth and block properties (OpenGL 4.6 "8.7. COMPRESSED TEXTURE IMAGES" p. 228)

I read this section and it's really confusing.

I'll need to think a bit more, but IIUC the solution is to:
* compute `imageOffset = UNPACK_SKIP_PIXELS/bw*n + UNPACK_SKIP_ROWS/bh*k`
* malloc(imageOffset + imageSize)
* zero the first imageOffset bytes
* copy the following imageSize bytes
* write a blob with the malloc contents.

Because essentially the spec mandates that the first `imageOffset` bytes should be skiped, and not touched.  Hence they might be backed with virtual address pages at all.

We can also compute the expected `imageSize` value, and either take the larger or the shorter of the two.  Not sure what would be the best.

We're definitely going to need a sample/test for this in https://github.com/apitrace/apitrace-tests .

werman:
> > The spec clearly states that imageSize depends only on width, height, depth and block properties (OpenGL 4.6 "8.7. COMPRESSED TEXTURE IMAGES" p. 228)
> 
> I read this section and it's really confusing.
> 
> I'll need to think a bit more, but IIUC the solution is to:
> 
>     * compute `imageOffset = UNPACK_SKIP_PIXELS/bw*n + UNPACK_SKIP_ROWS/bh*k`
> 
>     * malloc(imageOffset + imageSize)
> 
>     * zero the first imageOffset bytes
> 
>     * copy the following imageSize bytes

Small correction - it's not possible to just take `imageSize` consecutive bytes since we are picking a rectangle inside the whole image:

image
 

> 
>     * write a blob with the malloc contents.
> 
> 
> Because essentially the spec mandates that the first `imageOffset` bytes should be skiped, and not touched. Hence they might be backed with virtual address pages at all.

I thought that touching these bytes wouldn't be that bad, but yeah I agree it could go wrong.

> 
> We can also compute the expected `imageSize` value, and either take the larger or the shorter of the two. Not sure what would be the best.

From what I saw e.g. Mesa just goes with its own calculations based on unpack parameters since `imageSize` bytes may not be in a block of continuous memory.

Ultimately I can follow the way Mesa copies subimages.

> 
> We're definitely going to need a sample/test for this in https://github.com/apitrace/apitrace-tests .

After looking at VK-GL-CTS and Piglit I think such tests should also go there because this part of compressed images aren't tested at all from what I saw.

werman:
And there is also depth which handled even more confusedly  in spec, I still don't understand how UNPACK_SKIP_IMAGES, depth and sub-rectangles should interact. The most sane thing would be if it is like cube inside a cube. But I need to look with fresh at it.

jrfonseca:
After more thinking I think that mimicking what (correct) OpenGL drivers do internally seems a reasonable thing to do.

I'd really like to have a basic test for this.

Another showstopping issue I just noticed is OpenGL ES support -- I fear this PR as-is might break OpenGL ES support where perhaps not all used `glGetIntegerv(...)` are supported, no?

werman:
> After more thinking I think that mimicking what (correct) OpenGL drivers do internally seems a reasonable thing to do.
> 
> I'd really like to have a basic test for this.

Sorry that it takes so long. Since there are no tests for this functionality anywhere, we want to bring them to the test suits first. Currently there is MR pending into Piglit https://gitlab.freedesktop.org/mesa/piglit/merge_requests/193. Also we requested a small change in GL spec to clarify one ambiguity https://github.com/KhronosGroup/OpenGL-API/issues/62#issuecomment-573644092.

So the current plan is to have the test merged into Piglit, then we'll port them here and update the MR.

> Another showstopping issue I just noticed is OpenGL ES support -- I fear this PR as-is might break OpenGL ES support where perhaps not all used `glGetIntegerv(...)` are supported, no?

That's an interesting question. In es 2.0 spec these enums are not directly mentioned, on the other hand https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glPixelStorei.xhtml says that it should be supported in 2.0. However on the worst case `GL_INVALID_ENUM` error would be generated and we can just return `imageSize` in such case.

jrfonseca:
> So the current plan is to have the test merged into Piglit, then we'll port them here and update the MR.

Great.  Thanks.

> However on the worst case `GL_INVALID_ENUM` error would be generated and we can just return `imageSize` in such case.

I think it's too dangerous to rely on `glGetError` when tracing, as it might cause unintended consequences and disrupt the app.

But I see now there's a `can_unpack_subimage` parameter, which does the right thing.

jrfonseca:
Sorry for the delay.

I've merged this PR as is for now.  We can deal with https://github.com/apitrace/apitrace/pull/658 independently.

PR #631, opened 2019-10-13
Support tracing PBOs on GLES 3.0

Daft-Freak:
Just decrease the version check. Fixes a segfault tracing an app that uses PBOs on a Raspberry Pi 4.


jrfonseca:
Are you sure PBOs are part of GLES 3.0?  I'll double check but I thought I'd check this right

jrfonseca:
You're quite right! My bad.

Issue #633, opened 2019-10-30
Civilization VI traces corrupt


janesma:
Traces made of Civ6 fail to parse on a call to glCompressedTexImage2D

I've reproduced this with master and the 8.0 tag.

To capture the trace, I set launch options in steam: 
```
  {path_to}/apitrace trace -o /tmp/civ.trace %command%
```
You can get the parse error with
```
   > apitrace dump /tmp/civ.trace
   error: (glCompressedTexImage2D) unknown call detail 81
```
After parsing the blob from param 7 of glCompressedTeximage2D, the following byte is not a call detail.  Several calls to glCompressedTexImage2D successful before the crash.

I can't figure out how to debug the tracing through the steam engine, but I would be happy to look if someone can point me in the right direction.

A sample trace file is at:

 http://swizzler.org/imgcache/civ.trace.xz

werman:
Could you try with https://github.com/apitrace/apitrace/pull/626?

jrfonseca:
@werman appologies. I'll try to get it commited regardless.

janesma:
Thanks for the pointer!  It definitely looks like the type of issue that would cause my crash.  Unfortunately, I couldn't get Civ 6 to trace after building apitrace master with the fix from #626.

werman:
Additional info about launching the game, add:  
`LD_PRELOAD=/usr/lib/libfreetype.so.6 LD_LIBRARY_PATH=/usr/lib/openssl-1.0` for game to launch, at least on arch.
To bypass launcher (it is stuck when being traced) use:
`~/.steam/steam/steamapps/common/Sid\ Meier\'s\ Civilization\ VI/Civ6Sub %command%`

werman:
Now I'm unable to reproduce the issue.

janesma:
I can still reproduce the issue when I try to create a trace with the launcher options method.  However, I have found a method that does take a working trace:

```
cd ~/.steam/steam/steamapps/common/Sid\ Meier\'s\ Civilization\ VI
~src/apitrace/build/apitrace trace -o /tmp/civ.trace /home/majanes/.steam/ubuntu12_32/steam-runtime/run.sh ~/.steam//steam/steamapps/common/Sid\ Meier\'s\ Civilization\ VI/Civ6Sub -benchmark
```

For some reason, the launch options method for taking traces no longer works for several steam titles.  My steam client was updated, which may have something to do with it.  Fewer windows are shown when launching from the command line.

[...]

PR #634, opened 2019-11-04
trace: Fix uint serialization

werman:
`_writeUInt` wrongly serialized certain numbers: 255, 383, ...

Closes: #633

See http://cpp.sh/5w3av for example reproducing the issue.

### DIFF
--- a/lib/trace/trace_writer.cpp
+++ b/lib/trace/trace_writer.cpp
@@ -102,7 +102,7 @@ Writer::_writeByte(char c) {
 
 void inline
 Writer::_writeUInt(unsigned long long value) {
-    char buf[2 * sizeof value];
+    unsigned char buf[2 * sizeof value];
     unsigned len;
 
     len = 0;


jrfonseca:
@werman , I looked at the code, and didn't see how it could matter.

I also tried it locally with the numbers you mentioned, and it was working without your change:

    diff --git a/lib/trace/trace_parser.cpp b/lib/trace/trace_parser.cpp
    index 73e287c8..94c48c40 100644
    --- a/lib/trace/trace_parser.cpp
    +++ b/lib/trace/trace_parser.cpp
    @@ -59,6 +59,7 @@ bool Parser::open(const char *filename) {
         }
     
         version = read_uint();
    +    std::cerr << "version = " << version << "\n";
         if (version > TRACE_VERSION) {
             std::cerr << "error: unsupported trace format version " << version << "\n";
             delete file;
    diff --git a/lib/trace/trace_writer.cpp b/lib/trace/trace_writer.cpp
    index 1334a0eb..096e47b3 100644
    --- a/lib/trace/trace_writer.cpp
    +++ b/lib/trace/trace_writer.cpp
    @@ -76,10 +76,16 @@ Writer::open(const char *filename,
         bitmasks.clear();
         frames.clear();
     
    -    _writeUInt(TRACE_VERSION);
    +    //_writeUInt(TRACE_VERSION);
    +    _writeUInt(383);
    +    os::log("apitrace: warning: 255\n");
     
         assert(semanticVersion <= TRACE_VERSION);
         _writeUInt(semanticVersion);
    +    
    +    os::log("apitrace: warning: XXX\n");
    +    close();
    +    os::abort();
     
         beginProperties();
         for (auto & kv : properties) {
    diff --git a/lib/trace/trace_writer_local.cpp b/lib/trace/trace_writer_local.cpp
    index ee4d3a82..ce3b229a 100644
    --- a/lib/trace/trace_writer_local.cpp
    +++ b/lib/trace/trace_writer_local.cpp
    @@ -144,6 +144,7 @@ LocalWriter::open(void) {
         os::String processName = os::getProcessName();
         properties["process.name"] = processName;
     
    +        os::log("apitrace: YYY\n");
         if (!Writer::open(lpFileName, TRACE_VERSION, properties)) {
             os::log("apitrace: error: failed to open %s\n", lpFileName);
             os::abort();

I also have troubles believing this wasn't working,  we'd see lot of bugs like now.

Can you double check your findings?

werman:
After triple checking I found that `Parser::read_uint` actually reads unsigned chars which makes it correct. So the patch shouldn't fix anything.

I'm unable to reproduce the issue on my home pc which has too different hardware than my work laptop where I reproduced it. I'll double check there on Monday.

jrfonseca:
I used a `char[]` buffer because that's what `std::ostream::write` takes as parameter.  But at that point, is just a bunch of bytes to be written to disk (the signedness shouldn't matter.)

Do reopen if you find anything.  But the only this could make a difference is if it works around some mysterious compiler bug.

PR #636, opened 2019-11-12
glws_waffle: Add egl surfaceless support

afrantzis:
Bump the waffle version requirement to 1.6 and handle the new platform
token for EGL surfaceless.

EGL surfaceless is useful when running in a headless setup, typically as
part of some CI system.

Note that for eglretrace to work correctly with EGL surfaceless a [recent](https://gitlab.freedesktop.org/mesa/waffle/commit/8a5f50234b171d227796571b8cb2c882f7a2f26f)
version of waffle is required. 

Signed-off-by: Emil Velikov <emil.velikov@collabora.com>
Signed-off-by: Alexandros Frantzis <alexandros.frantzis@collabora.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ endif ()
 
 if (ENABLE_EGL AND ENABLE_WAFFLE)
     # Use Waffle for eglretrace
-    find_package (Waffle REQUIRED)
+    find_package (Waffle 1.6 REQUIRED)
 endif ()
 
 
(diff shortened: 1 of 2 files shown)


jrfonseca:
I wonder if we should make Waffle 1.6 requirement conditional, and only enable surfaceless EGL only when Waffle version >= 1.6.  Otherwise looks good.

For now, I'll merge this as-is.  We can make Waffle 1.6 conditional later if somebody complains about the new requirement.

Thanks.

PR #637, opened 2019-11-15
Allow building without X11 on __unix__

glebm:
eglplatform.h updated from https://www.khronos.org/registry/EGL/api/EGL/eglplatform.h.

Previously, `eglplatform.h` tried to include `#include <X11/Xlib.h>` even with `-DENABLE_X11=off`.


jrfonseca:
Thanks.  I had to update `EGL/eglplatform.h` and commit seperately, because there were some modifications that got lost.

PR #638, opened 2019-11-19
dlltrace: Implement dll path override

Joshua-Ashton:
Implement env var APITRACE_FORCE_MODULE_PATH which allows the user to force override the directory to the dirs we want to trace against.

This is useful for tracing against reimplementations (ie. d9vk/dxvk) on their native platforms.

### DIFF
--- a/wrappers/dlltrace.hpp
+++ b/wrappers/dlltrace.hpp
@@ -28,6 +28,8 @@
 
 #include <windows.h>
 #include <string>
+#include <cstdlib>
+#include <cstring>
 
 
 class Module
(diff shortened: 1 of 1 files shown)


wrappers/dlltrace.hpp
```diff
@@ -28,6 +28,8 @@
 
 #include <windows.h>
 #include <string>
+#include <cstdlib>
```
jrfonseca: I prefer we don't use C++ variants of C headers, because they are often unreliable.  Otherwise looks good.

jrfonseca:
Pushed.

Issue #142, opened 2013-06-08
glretrace --loop should not be snappy or parse limited


casseveritt:
I'd like to --loop the last frame in playback and see an fps output that is not limited by glretrace CPU usage.  When I run a profiler in this steady state mode, the top 20 or so functions are all in the snappy library.

Making this work may be as simple as keeping around in memory that last frame's worth of Call objects and just sequencing through them repeatedly.  That will still have the overhead of mapping record names and uniform slots to playback names and slots, but maybe that won't be a performance limiter.  It's pretty clear that snappy decompression is limiting playback performance right now though.

Alternatively, maybe there's a way to unpack a trace so it's not in the loop for these kinds of operations?

jrfonseca:
I thought about this again, and I believe the ultimate solution is to use LLVM to JIT compile each call into x86 code, and then call them all in succession.

This would be an excellent student project if anybody would be interest.

janesma:
I was able to work around the CPU bottlneck during parse by distributing the parse work across multiple threads.  My POC is a hack, because the parser relies on state that is built when the trace is parsed sequentially (eg FunctionSigState) .  A parser needs to scan the file once, setting bookmarks at intervals that can be delegated to an off-thread parser for generating Call objects.

The pre-pass is time consuming, but it allows a subsequent run to utilize all cores of the hardware during parse, with a speedup that is linear for some traces.  For my target hardware (gen9 Intel GPU) a threaded retrace uses the full throughput of the GPU, making apitrace suitable for driver benchmarking.

janesma:
I think this issue can be closed now that #639 has been merged.

eero-t:
> I think this issue can be closed now that #639 has been merged.

I think support for leaner traces thing mentioned above is also important (I didn't see any other ticket mentioning it), and there may be some smaller potential wins too (e.g. optimizing some of the alloc paths listed in #146).

PR #639, opened 2019-11-26
retrace: when looping on last frame, re-use Call objects

janesma:
Call objects are unnecessarily parsed when looping on the last frame.
Parsing incurs significant CPU overhead, often limiting performance
and making Apitrace unusable for benchmarking.

https://github.com/apitrace/apitrace/issues/142


jrfonseca:
Looks good to me.  Thanks.

PR #644, opened 2019-12-06
dlsym: Fix build under uClibc

pcercuei:
uClibc defines the ```__GLIBC__``` macro too (I know, right?), so verify that
```__UCLIBC__``` is not defined before we can use glibc-specific internal
functions.

Signed-off-by: Paul Cercueil <paul@crapouillou.net>


jrfonseca:
Thanks

PR #646, opened 2019-12-13
Support libc that don't have RTLD_DEEPBIND

glebm:
Same as https://github.com/apitrace/apitrace/commit/508a26ab7d613a82d8024309582dd34f400ded8e but for `glproc_gl.cpp`


jrfonseca:
Thanks.

PR #648, opened 2020-01-02
Show depth for 3D textures

10110111:
This PR changes the "x" letter used instead of multiplication sign in texture sizes (e.g. _16 x 32_ instead of _16 × 32_), and adds the size of the third dimension, depth, for the textures for which it's relevant.


jrfonseca:
Thank you.

PR #649, opened 2020-01-06
Support cross-platform retrace for titles that query uniforms by index

janesma:
These patches enable Linux/Windows retrace for GFXBench AztecRuins and Bioshock Infinite.

I'm happy to re-write if there is a better way to do the code generation.

### DIFF
--- a/retrace/glretrace_main.cpp
+++ b/retrace/glretrace_main.cpp
@@ -996,3 +996,24 @@ glretrace::trackResourceName(GLuint program, GLenum programInterface,
         uniform_vec.resize(index + 1);
     uniform_vec[index] = traced_name;
 }
+
+void
+glretrace::mapUniformBlockName(GLuint program,
+                               GLint index,
+                               const std::string &traced_name,
+                               std::map<GLuint, retrace::map<GLuint>> &uniformBlock_map) {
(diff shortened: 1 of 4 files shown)


retrace/glretrace_main.cpp
```diff
@@ -996,3 +996,24 @@ glretrace::trackResourceName(GLuint program, GLenum programInterface,
...
+        std::vector<char> name_buf(buf_len+1);
+        GLint length;
+        glGetActiveUniformBlockName(program, i, name_buf.size(), &length, name_buf.data());
+        name_buf[length] = '\0';
+        if (traced_name == name_buf.data()) {
+            uniformBlock_map[program][index] = i;
+            return;
+        }
+    }
+}
```
jrfonseca: It might be useful to log an error here to debug any issues on the future.

janesma:
I will add some error handling.  I'm only handling uniform locations that are found via GL_ARB_program_interface_query, but there are other properties that may differ between drivers.  I didn't have a trace that broke for any reason other than uniform locations.

I'm looking at Shadow of Mordor now, which fails to re-map shader storage block bindings.  I'll add fix-up error handling in the pull request for that fix.

PR #650, opened 2020-01-19
Remove executable permission from .cpp files

akien-mga:
It's not harmful per se but also not necessary, and triggers warnings
e.g. when building RPM packages.


jrfonseca:
Thanks.

PR #651, opened 2020-01-22
glws_glx: Allow pixel buffers in GlxDrawable::swapBuffers

werman:
Pixel buffers are allowed to be used in glXSwapBuffers.

glx 1.4 spec, 3.3.5 "Off Screen Rendering", "glXCreatePbuffer":

 "The resulting pbuffer will contain color buffers and ancillary
  buffers as specified by config. It is possible to create
  a pbuffer with back buffers and to swap the front and back
  buffers by calling glXSwapBuffers."

Fixes replaying of Blender traces.

### DIFF
--- a/retrace/glws_glx.cpp
+++ b/retrace/glws_glx.cpp
@@ -150,7 +150,6 @@ public:
     }
 
     void swapBuffers(void) override {
-        assert(!pbuffer);
         if (window &&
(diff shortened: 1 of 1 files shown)


jrfonseca:
Good point.  Sorry it took this long to merge such trivial change.  I wrongly presumed it was something more complex.

PR #653, opened 2020-02-07
gltrace: Commit all writes to coherent buffers before syncing them back

werman:
Otherwise these writes would be lost. Deus Ex: Mankind Divided renders incorrectly without this.


jrfonseca:
Thanks

PR #655, opened 2020-03-18
glretrace: Added per frame delay options in replay

tutan-chromium:
Added two extra options to control frame rendering duration in replaying
mode. --per-frame-delay option adds extra delay after each frame in usec.
--min-frame-duration option specifies the minimum frame duration time in usec.


jrfonseca:
Thanks.  And sorry for the delay.

PR #658, opened 2020-03-24
gltrace: Consider unpack parameters when writing compressed texture

werman:
This makes https://github.com/apitrace/apitrace/pull/626 obsolete. 

It's a bit weird in how it handles the writes to a blob  - I hope it's not bad.

The `canTakeFastPath` function is similar to RenderDoc - https://github.com/baldurk/renderdoc/blob/af1f98039b262c85ec7426aeb82d6ac2a3f1dc42/renderdoc/driver/gl/gl_renderstate.cpp#L222

I still don't have apitrace tests for this, but it traces correctly Piglit tests we wrote for unpack parameters which were evolved into GL-CTS tests https://github.com/KhronosGroup/VK-GL-CTS/pull/195


batesj:
This doesn't seem to fix #626 -- with your patch, certain replays that use glCompressedTexSubImage2D still crash for me.

werman:
> This doesn't seem to fix #626 -- with your patch, certain replays that use glCompressedTexSubImage2D still crash for me.

Are you using old traces? This patch fixes recording of the compressed textures, therefor you should record traces with this commit applied for them to be replayed correctly.

If there is an issue even after recording a trace with this changes - I would like to know how to reproduce it.

batesj:
I see, yes it was an old trace. Unfortunately I can't easily redo this particular trace. I wonder if we should use a trace format version to catch such errors -- currently it will read outside the buffer on old traces, but if the version of the trace file was checked could the behavior be changed to either error out on use of glCompressedTexImage or to avoid the out of bounds reads?

werman:
It's probably doable to check if glCompressedTexImage is being replayed with unpack params set and skip it, however I don't think it's worth it. Also I don't know much about apitrace versioning so I would be a wrong person to answer.

werman:
The tests were merged in VK-GL-CTS, I my changes retested by:

```
apitrace trace ./glcts --deqp-case="KHR-GL33.pixelstoragemodes.*"  --deqp-log-images=disable --deqp-log-shader-sources=disable
```

Then I replayed the trace, and without this commit - there are non-green rectangles (test expects to output only green ones), with the commit - all rectangles are green.

@jrfonseca I know you asked for tests for apitrace-tests, but could the tests in GL-CTS be enough?

jrfonseca:
@werrman, sorry for the delay.

I've merged https://github.com/apitrace/apitrace/pull/626 as is for now, as it's been so long.  But I'm happy to revert and merge this instead.

@weman, @batesj, @sergiiromantsov, can anybody confirm this PR also fixes the same issue as https://github.com/apitrace/apitrace/pull/626 ?

(Sorry, I really don't have the time to do it myself, and I want to stop  stalling merging PRs until I have time, as I've been doing so far, as it benefits nobody.)

werman:
> @werrman, sorry for the delay.
> 
> I've merged #626 as is for now, as it's been so long. But I'm happy to revert and merge this instead.
> 
> @weman, @batesj, @sergiiromantsov, can anybody confirm this PR also fixes the same issue as #626 ?
> 
> (Sorry, I really don't have the time to do it myself, and I want to stop stalling merging PRs until I have time, as I've been doing so far, as it benefits nobody.)

I extensively tested this change with CTS and a few games. However quite some time have passed, it's better to recheck it. @sigexp Could you recheck it, please.

jrfonseca:
@sergiiromantsov if you're OK using this version instead, I'm OK too.

sigexp:
> @sigexp Could you recheck it, please.

I've checked it: indeed, traces of my tests in VK-GL-CTS, captured without this commit, show incorrect colours on replay. Similarly, on replay of Dota 2, there are artifacts all over the grass. With the commit, traces of both the game, and the tests, are displayed as expected.

sergiiromantsov:
@jrfonseca, I'm OK

jrfonseca:
Merged.  Thanks

PR #660, opened 2020-03-31
Add the new client application 'info'

tutan-chromium:
The new 'info' client prints the provided trace file information. The
first version of the client shows the container version, api type and
the total amount of frames in the trace file. The information can be
printed in JSON format.


jrfonseca:
Looks useful.  Thanks.

PR #665, opened 2020-05-23
gui: fix keyboard focus to make hotkeys work in image viewer

ChipmunkV:
There are zoom hotkeys (+/-), grid hotkeys (g, PgUp, PgDown), etc that don't seem to work when viewing an image of a texture. This change allows PixelWidget to get keyboard focus.


jrfonseca:
LGTM.  Thanks

PR #666, opened 2020-05-23
gui: make image viewer windows maximizable

VelorumS:
ImageViewer is a dialog window with no maximizing ability (double-click on the header of the window doesn't maximize, there is no maximize button). So making it a proper window.

Side effect with this patch is that the viewer windows can be behind the main window.

### DIFF
--- a/gui/imageviewer.cpp
+++ b/gui/imageviewer.cpp
@@ -15,6 +15,7 @@ ImageViewer::ImageViewer(QWidget *parent, bool opaque, bool alpha)
       m_image(0)
 {
     setupUi(this);
+    setWindowFlags(Qt::Window);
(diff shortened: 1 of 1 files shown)


jrfonseca:
Neat.  Thanks

jrfonseca:
@ChipmunkV, you marked this as draft/work-in-progress. Was that intentional?

ChipmunkV:
@jrfonseca yes, it's a draft. Since I can't make them stay on top reliably it's not convenient (there is maybe a hack that works on some systems).

So the better thing to do is to keep it a dialog like before, but bind the maximize action (if it works for the dialogs) to the double-click on the image itself like the youtube videos or all image viewers do.

PR #667, opened 2020-05-23
gui: preserve main window size and layout between restarts

ChipmunkV:
Save main window dimensions/maximized, layout of dock elements.


jrfonseca:
Thanks,

PR #669, opened 2020-05-28
gui: zoom to fit in image viewer

ChipmunkV:
Center the image, focus PixelWidget at start to make keys work, allow zoom < 1, add a checkbox to auto-fit the image (stops fitting when manually zoomed), save auto-fit option after restart.


jrfonseca:
I tried it, and it looks great!

I've merged it.  Sorry it took this long.

Thanks

PR #670, opened 2020-06-02
glapi: Add glSpecializeShaderARB from GL_ARB_gl_spirv

rdb:
Thank you so much for maintaining this tool!  It's an absolute godsend for debugging any kind of graphics issue in our engine.

This change makes it possible for apitrace to work with applications that use SPIR-V shaders.  Without this, glLinkProgram would fail and glretrace would crash.

Thanks for your consideration!


specs/glapi.py
```diff
@@ -1228,6 +1228,9 @@ def OutGlString(charType, lengthPtr, argName):
     GlFunction(Void, "glGetTextureSubImage", [(GLtexture, "texture"), (GLint, "level"), (GLint, "xoffset"), (GLint, "yoffset"), (GLint, "zoffset"), (GLsizei, "width"), (GLsizei, "height"), (GLsizei, "depth"), (GLenum, "format"), (GLenum, "type"), (GLsizei, "bufSize"), Out(GLpointer, "pixels")]),
     GlFunction(Void, "glGetCompressedTextureSubImage", [(GLtexture, "texture"), (GLint, "level"), (GLint, "xoffset"), (GLint, "yoffset"), (GLint, "zoffset"), (GLsizei, "width"), (GLsizei, "height"), (GLsizei, "depth"), (GLsizei, "bufSize"), Out(GLpointer, "pixels")]),
 
+    # GL_ARB_gl_spirv
+    GlFunction(Void, "glSpecializeShaderARB", [(GLuint, "shader"), (GLstringConst, "pEntryPoint"), (GLuint, "numSpecializationConstants"), (Array(Const(GLuint), "numSpecializationConstants"), "pConstantIndex"), (Array(Const(GLuint), "numSpecializationConstants"), "pConstantValue")]),
```
jrfonseca, line 1232: I think it should be `GLshader, "shader"` instead of `GLuint, "shader"`  Otherwise looks good.

jrfonseca: On the other hand `glShaderBinary` uses `uint` instead of `GLshader` so I'm less sure.

I'll commit as is, and we can tweak later.

rdb: Thanks a lot for merging!

PR #671, opened 2020-06-03
gltrace: Forward unhandled signals/exceptions to a next handler

werman:
Some applications have custom handler for signals/exceptions,
which is necessary for their working. Therefor we shall forward
all unhandled signals/exceptions down the road.

Also apitrace itself has custom logic for handling signals,
which failed to execute due to lack of signal forwarding in
GLMemoryShadow.

Fixes a crash when tracing rpcs3.

Fixes: 18ce32f949409d051c88aea62209818aec7d6aa5


jrfonseca:
Some of the details escape me, but makes sense.  Thanks, and sorry for the delay.

PR #674, opened 2020-06-26
Avoid exponential notation when showing binary data size

10110111:
Also add a thin no-break space between the number and the unit, and use
a larger set of units, with explicitly-binary prefixes: i.e. KiB, MiB, GiB.


jrfonseca:
Looks good to me.  Thanks

PR #678, opened 2020-07-24
Fix glMultiDrawArraysIndirect not capturing client buffer

Sturmlilie:
I'm not entirely sure why using `Blob` instead of `OpaqueArray` makes this work, but it does.
In case `stride` is zero, it means "the array is tightly packed" [1].

[1] https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_multi_draw_indirect.txt

### DIFF
--- a/helpers/glsize.hpp
+++ b/helpers/glsize.hpp
@@ -1267,3 +1267,13 @@ _glGetDebugMessageLog_length(const Char * string, const GLsizei *lengths, GLuint
     return size;
 }
 
+static inline size_t
+_glMultiDrawArraysIndirect_size(GLsizei drawcount, GLsizei stride)
+{
+    if (stride == 0) {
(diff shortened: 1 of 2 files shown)


Sturmlilie:
I just realized that the command allows (and is primarily geared towards) usage with a buffer object bound, in which case this patch will likely copy from an invalid memory address. Will investigate how to differentiate between the two cases.

jrfonseca:
It's not easy to do this sort of stuff (polymorphism) just from the spec, but it can be done.

See https://github.com/apitrace/apitrace/blob/15de87464417ffb072fea4f7f0ea2cec119cc3fc/specs/gltypes.py#L305 for example, on how the index buffers are handled.  They too can either be a memory pointer, or a buffer offset, depending on whether an index buffer is bound.

PR #681, opened 2020-08-10
glws_glx: Choose pbuffer's attribs according to visual info

werman:
Otherwise we may get a pbuffer with configuration incompatible
with a context.

Fixes replaying of traces of Substance Painter.
Found in: https://gitlab.freedesktop.org/mesa/mesa/-/issues/3271

Trimmed trace which reproduces the crash (at least on Mesa): 
[substance_painter_trimmed.zip](https://github.com/apitrace/apitrace/files/5055602/substance_painter_trimmed.zip)

### DIFF
--- a/retrace/glws_glx.cpp
+++ b/retrace/glws_glx.cpp
@@ -446,21 +446,29 @@ GlxDrawable::createPbuffer(Display *dpy, const GlxVisual *visinfo,
 {
     int samples = 0;
     int doubleBuffer = 0;
+    int red_size = 0, green_size = 0, blue_size = 0, alpha_size = 0;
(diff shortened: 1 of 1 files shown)


jrfonseca:
It's not easy to do this sort of stuff (polymorphism) just from the spec, but it can be done.

See https://github.com/apitrace/apitrace/blob/15de87464417ffb072fea4f7f0ea2cec119cc3fc/specs/gltypes.py#L305 for example, on how the index buffers are handled.  They too can either be a memory pointer, or a buffer offset, depending on whether an index buffer is bound.

jrfonseca:
Looks good to me.  Thanks.

PR #686, opened 2020-09-23
retrace: Print frame times

Fahien:
Hello there,

The commit messages should be explicative enough:

```
When replaying a trace, it could be useful to analyze how much time is
needed to render and present each frame. While developing a GPU driver,
for example, developers could graph the frame time series in order to
detect regressions or improvements of driver performance over time.

The `--pframe-times` option enables profiling for frame times, meaning
that replaying will output a timestamp when a frame has been presented.
```

I guess this is a first draft of something that needs some more work.
At the current state, this option works only on `wgl`, `glx`, and `egl` backends.
Please, tell me what I am missing, and how I could tackle that. 😃

### DIFF
--- a/docs/USAGE.markdown
+++ b/docs/USAGE.markdown
@@ -394,6 +394,8 @@ section above.
 
 You can perform gpu and cpu profiling with the command line options:
 
+ * `--pframe-times` record frames times, waiting for presentation to finish before starting next frame.
+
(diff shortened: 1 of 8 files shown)


tomeuv:
@jrfonseca Is it looking like you would have time soon to review this? Thanks!

jrfonseca:
Looks good.  Thanks.  Sorry for the delay

Issue #433, opened 2016-03-03
Support for minimizing traces through delta debugging


sesse:
Hi,

I recently used apitrace to make a reproducible test case for an OpenGL driver crashing. (It's fantastic for that!) However, the test case was ~8000 calls, which made it hard to see exactly what was going on, and trim wasn't much help. It struck me that this is a great match for delta debugging, a standardized algorithm for trying to minimize deltas or test cases given automated testing.

I wrote a module for DD.py (https://www.st.cs.uni-saarland.de/dd/DD.py) that splits the trace file into individual calls and then lets DD.py remove calls in a systematic fashion until only the bare minimum was left. It's super-effective; here's what it ended up with:

```sesse@gruessi:~/nmu/dd$ apitrace dump nageru.2.trace 
0 glXCreateContextAttribsARB(dpy = 0x1dddc60, config = 0x1e4ca10, share_context = 0x23a8350, direct = True, attrib_list = {GLX_CONTEXT_MAJOR_VERSION_ARB, 3, GLX_CONTEXT_MINOR_VERSION_ARB, 1, GLX_CONTEXT_FLAGS_ARB, GLX_CONTEXT_FORWARD_COMPATIBLE_BIT_ARB, 0}) = 0x7f6ed8001670
1 glXMakeCurrent(dpy = 0x1dddc60, drawable = 20971619, ctx = 0x7f6ed8001670) = True
2 glTexImage2D(target = GL_TEXTURE_2D, level = 0, internalformat = GL_R8, width = 720, height = 525, border = 0, format = GL_RED, type = GL_UNSIGNED_BYTE, pixels = NULL)
3 glTexSubImage2D(target = GL_TEXTURE_2D, level = 0, xoffset = 0, yoffset = 0, width = 360, height = 525, format = GL_RG, type = GL_UNSIGNED_BYTE, pixels = blob(378000))
4 glXCreateContextAttribsARB(dpy = 0x1dddc60, config = 0x1e4ca10, share_context = 0x23a8350, direct = True, attrib_list = {GLX_CONTEXT_MAJOR_VERSION_ARB, 3, GLX_CONTEXT_MINOR_VERSION_ARB, 1, GLX_CONTEXT_FLAGS_ARB, GLX_CONTEXT_FORWARD_COMPATIBLE_BIT_ARB, 0}) = 0x7f6eb8001640
[...]

PR #691, opened 2020-11-17
Gltrim

gerddie:
This PR adds support for a new trimming tools *gltrim* that trims OpenGL/GLES traces to a given series of frames. Other then the *trim* command already available, this tools aims at keeping the trimmed trace replayable. The resulting trace consists of one setup frame that contains all calls required to properly replay the frames of interest. 

The tool does not aim to be a solution that works fully automated on all traces, for this OpenGL offers just too many possibilities to let textures and buffers depend on each other. For that it is possible to specifying additional frames that are considered to belong to the setup who's calls are added to the output setup frame (without the frame-end call). Additionally one can force that all simple state calls are kept in the trace. 

The tool has been tested on traces of various games, amongst them *Portal 2*, *Stellaris*, *Alien Isolation*, *Civilization V*, *Unturned*, and *Soma*.  Here for the used traces *Portal 2* and *Stellaris* could be trimmed to a single replayable frame without any problem. For *Alien Isolation* it was required to keep all state changes, *Unturned* needed one additional setup frame, and both  *Civilization V*, and *Soma* required to keep additional frames directly before the target frame because they build output over various frames. 

The pull request is considered to be a draft because not all OpenGL functions are supported yet.

### DIFF
--- a/frametrim/ft_dependecyobject.hpp
+++ b/frametrim/ft_dependecyobject.hpp
@@ -1,0 +1,242 @@
+/*********************************************************************
+ *
+ * Copyright 2020 Collabora Ltd
+ * All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use, copy,
+ * modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *********************************************************************/
+
+#pragma once
+
+#include "ft_tracecall.hpp"
+
+namespace frametrim {
+
+class UsedObject {
+public:
+    using Pointer = std::shared_ptr<UsedObject>;
+
+    UsedObject(unsigned id);
+
+    unsigned id() const;
+
+    void addCall(PTraceCall call);
+    void setCall(PTraceCall call);
+
+    void addDependency(Pointer dep);
+    void setDependency(Pointer dep);
+
+    void emitCallsTo(CallSet& out_list);
+    bool emitted() const;
+
+    unsigned extraInfo(const std::string& key) const;
+    void setExtraInfo(const std::string& key, unsigned value);
+
+private:
+
+    std::vector<PTraceCall> m_calls;
+    std::vector<Pointer> m_dependencies;
+    unsigned m_id;
+    bool m_emitted;
+    std::unordered_map<std::string, unsigned> m_extra_info;
+};
+
+class DependecyObjectMap {
+public:
+    using ObjectMap=std::unordered_map<unsigned, UsedObject::Pointer>;
+
+    void generate(const trace::Call& call);
+    void destroy(const trace::Call& call);
+
+    void create(const trace::Call& call);
+    void del(const trace::Call& call);
+
+    UsedObject::Pointer bind(const trace::Call& call, unsigned obj_id_param);
+    void callOnBoundObject(const trace::Call& call);
+    UsedObject::Pointer bindWithCreate(const trace::Call& call, unsigned obj_id_param);
+    void callOnObjectBoundTo(const trace::Call& call, unsigned bindpoint);
+    void callOnNamedObject(const trace::Call& call);
+    UsedObject::Pointer
+    callOnBoundObjectWithDep(const trace::Call& call,
+                             DependecyObjectMap& other_objects, int dep_obj_param, bool reverse_dep_too);
+
+    void
+    callOnBoundObjectWithDepBoundTo(const trace::Call& call, DependecyObjectMap& other_objects,
+                                    int bindingpoint, CallSet &out_set, bool recording);
+
+    void callOnNamedObjectWithDep(const trace::Call& call,
+                                  DependecyObjectMap& other_objects, int dep_obj_param, bool reverse_dep_too);
+    void callOnNamedObjectWithNamedDep(const trace::Call& call,
+                                       DependecyObjectMap& other_objects,
+                                       int dep_call_param, CallSet& out_set, bool recording);
+
+    UsedObject::Pointer getById(unsigned id) const;
+    void addCall(PTraceCall call);
+
+    void emitBoundObjects(CallSet& out_calls);
+    UsedObject::Pointer boundTo(unsigned target, unsigned index = 0);
+
+    ObjectMap::iterator begin();
+    ObjectMap::iterator end();
+
+protected:
+
+    UsedObject::Pointer bind(unsigned bindpoint, unsigned id);
+    void addObject(unsigned id, UsedObject::Pointer obj);
+    UsedObject::Pointer boundAtBinding(unsigned index);
+
+private:
+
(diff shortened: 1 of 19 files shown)


jrfonseca:
I'll need more time to review and think.

We used to have ability to trim GL traces.  But it was so limited it never worked in practice.  So it was removed.  I fear if I add this it will end up removed later.  See f99bf0db20e0bffa6a69ee0f47f131fbddbcadaf and https://lists.freedesktop.org/archives/apitrace/2017-April/001215.html

Personally I think a tool based off the ideas described on https://github.com/apitrace/apitrace/issues/433 are much better.

gerddie:
Well, I'm aware of the old version and also of the approach in #433. With #433 I'd expect very long run-times, and defining a proper metric seems not to be straightforward for some traces. This tool here is quite fast, so the turnaround time when the result is not satisfying is very low.  

I think it all depends on the idea of what is one trying to achieve. The major use case here is supposed to be able to remove all the initial screens that games usually have to be able to render one or more frames of the actual game for further analysis. The tools is not expected to be immediately successful in 100% of the cases, there may even be cases where it will always fail. 
One application we are aiming at is using the tool in the Mesa CI to check rendering correctness with a few traces, and to bring the processing time down we would like to render just one frame. The alternative for this use case, renderdoc, doesn't and probably never will support compatibility contexts, that's the reason  why we thought to base this work on apitrace. 

To give you and idea of how well or bad it works: I've  testes on 14 different games of different age and using different game engines, out of which 10 worked out of the box by just specifying the target frame, two needed additional, hand picked, setup frames, one required the states to be kept, and with one game is probably not possible to always get a correct result with a reasonable number of setup frames because it uses motion blur by accumulating rendering results over time, and it is cases like this why I think that always getting everything right for just one frame is not the aim here,

Anyway, if you think that the danger is too hight, that people might consider that it should be removed at a later point, it could be maintained as a stand alone tool. For that we would need apitrace to install libraries and headers that allow to call the trace reading and writing, and I could propose patches for that instead.

gerddie:
BTW: as I pointed out in the gitlab issue, I will have time to help maintaining apitrace and also on this code.

frametrim/ft_dependecyobject.hpp
```diff
@@ -0,0 +1,220 @@
...
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *********************************************************************/
+
+#ifndef FT_DEPENDECYOBJECT_HPP
```
jrfonseca, line 28: This is OK as is, but for future reference I prefer `#pragma once`, as it's widely supported and less boilerplate code.

frametrim/tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    /bin/bash ${CMAKE_CURRENT_SOURCE_DIR}/compare-trace.sh
+    ${CMAKE_BINARY_DIR}
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${name}.test_trace
+    ${frame}
+    ${orig_img})
+ENDMACRO(TRACE_TEST)
+
+
+TRACE_TEST(glxgears 238 6620)
```
jrfonseca, line 14: These traces are quite big.  I've been refraining from adding large traces as test cases.

First, double check you recompress with brotli (ie, `aptirace repack -b ...`) to reduce their size.  If you really need a git repos, then we should create a sub repos.  But it might be better to find different kind of storage (ie out of tree.)

frametrim/ft_main.cpp
```diff
@@ -0,0 +1,278 @@
...
+        std::cerr << "error: extraneous arguments:";
+        for (int i = optind + 1; i < argc; i++) {
+            std::cerr << " " << argv[i];
+        }
+        std::cerr << "\n";
+        usage();
+        return 1;
+    }
+
+    return trim_to_frame(argv[optind], options);
```
jrfonseca, line 277: This utility should have a warning stating there are some limitations, perhaps pointing to `frametrim.markdown`

frametrim/ft_main.cpp
```diff
@@ -0,0 +1,278 @@
...
+            options.keep_all_states = true;
+            break;
+        default:
+            std::cerr << "error: unexpected option `" << (char)opt << "`\n";
+            usage();
+            return 1;
+        }
+    }
+
+    /*if (options.frames.getFirst() != options.frames.getLast())  {
```
jrfonseca, line 256: Remove

frametrim/ft_main.cpp
```diff
@@ -0,0 +1,278 @@
...
+const static char *
+shortOptions = "t:hko:f:s:x";
+
+bool operator < (std::pair<unsigned, unsigned>& lhs, std::pair<unsigned, unsigned>& rhs)
+{
+    return lhs.first < rhs.first;
+}
+
+const static struct option
+        longOptions[] = {
```
jrfonseca, line 92: Please indent

jrfonseca (commented):
Comments inline.

To be honest, I stil have some doubts this tool will stand the test of time, but if it works for you / your users and you're willing to maintain it, I won't object get it committed.

I only glance through the code and don't see anything bad.

I think the major blocker ire the binary trace files.  I don't want to weigh down apitrace git repos with lots of large blobs.  Perhaps a simple solution is to split out the trace test based from the PR, get it commited without them, and then address the addition of the test cases  on a subsequent PR.

It would be nice to have a shared repo with .trace files for testing (this and other stuff.)  I don't know exactly what's ideal, but think git is not it.

frametrim/ft_dependecyobject.hpp
gerddie: Changed.

frametrim/tests/CMakeLists.txt
gerddie: I've added a sub-repo and an option to enable the tests so that there is no need to always check out the tests.

frametrim/ft_main.cpp
gerddie: I've added a message when the program is run that points out that one should check the result and hinted to the markdown file for further details.

frametrim/ft_main.cpp
gerddie: Comment removed.

frametrim/ft_main.cpp
gerddie: Fixed.

jrfonseca:
Hi @gerddie,

Thanks for the update.

- let's pick a name `frametrim` or `gltrim` and use it consistently everywhere (filenames, symbols, directories, repos. etc.)

- We'll want https://github.com/gerddie/gltrim-tests.git  on `https://github.com/apitrace/gltrim-tests.git` .  I'll get it mirrored and give access to you next week (I'm PTO this week, taking care of other stuff.)   Also, I wonder if it would make sense to use https://docs.github.com/en/free-pro-team@latest/github/managing-large-files/configuring-git-large-file-storage / https://git-lfs.github.com/  .  I don't have any experience git LFS myself -- I just wonder if it would be adequate here.

- Are C++17 features worth it?  (See https://ci.appveyor.com/project/jrfonseca/apitrace/builds/36646968/job/g9ctp98xb3vmy8q3 ) 
 I think it's fine to start requiring C++17 everywhere sooner or later.  But if we want to do it now, this implies updating CMake and all CI build scripts to require C++17 everywhere.  That should be done independently and before this PR.

Otherwise all looks good to me.

gerddie:
(1) I settled on gltrim, but with the directory name `gltrim` I had a problem because  with that the executable could not be created that has the same name, so I opted for `trim` for the sub-directory name. 

(2) I would guess that using LFS  would be an alternative to using a submodule, but my experience is also limited 

(3) I've created a PR #694 to move requiring C++17 forward. Let's see whether the CI can handle it as is. I think it is worth updating,  C++17 has some nice features, and if you consider moving there anyway, we can do it also right now. It is probably less work that to rewrite the code that needs some c++17 features. 

I'll be away from 19.12. to 27.12.

gerddie:
@jrfonseca Thanks for adding me to the project. How do you want me to move forward here?

jrfonseca:
@gerddie , sorry for the silence, it's been difficult to make time for anything lately (school holidays, followed by stricter lockdown here in UK implying homeschooling), and there are some other PRs here also competing for my attention.

I've cloned https://github.com/apitrace/gltrim-tests  .  Please modify this PR to use it, and merge it yourself , at your leisure.  If there's any glitch after merging we can always deal with it.

Please use _rebase and merge_ if every commit is good by its own, or _merge commit_ if the commits don't make sense by themselves.  This makes it easier to do bisect.

gerddie:
@jrfonseca I've updated the submodule address, but it seems I don't have write access after all, so I can't merge. The commits should all be good on their own.

jrfonseca:
> but it seems I don't have write access after all

The new GitHub teams are way more complicated than it used to be.  Hopefully it's fixed now and you have commit access.

gerddie:
All good, thanks

PR #693, opened 2020-12-04
glstate: Fix out-of-bounds memset of image

werman:
E.g. the crash could be reproduced with this trace: [gl-3.2-layered-rendering-gl-layer-render.trace.zip](https://github.com/apitrace/apitrace/files/5642748/gl-3.2-layered-rendering-gl-layer-render.trace.zip)


jrfonseca:
Makes sense.  Thanks.

PR #694, opened 2020-12-09
Require C++17

gerddie:
As discussed in #691


jrfonseca:
I've just tweaked your change to require GCC 5.0, so it doesn't break on Ubuntu 16.04 (which we still use), and pushed it.  I'm sure we can get newer GCC on Ubuntu 16.04 by modifying the Dockerfile, but this is simpler for now.

Thanks.

PR #695, opened 2020-12-16
retrace: Fix GPU Duration

Fahien:
By running `eglretrace --pframes "opengl: GPU Duration" test.trace`, we would see a column with GPU time for each frame of the trace, but the actual value is incorrect.

That is because GPU duration relies on GPU start. If that is not enabled with "opengl: GPU Start", we end up using a uninitialized value, resulting in undefined behavior.

With this commit, we make sure GPU start is valid before calculating GPU duration.


jrfonseca:
I didn't write this code, so the fine details escape me, but sounds reasonable.  Thanks.

PR #696, opened 2020-12-18
specs: decode more info about ATI_fragment_shader operations

mtmkls:
Signed-off-by: Miklós Máté <mtmkls@gmail.com>


jrfonseca:
Thanks.  Looks good.  The missing `GLuint` args are bitfields.  If you're interested in handling those better you can see examples of other bitfields on `gltypes.py`.

PR #698, opened 2020-12-22
CMakeLists.txt: respect BUILD_TESTING=OFF

ffontaine:
Allow the user to disable unit tests through BUILD_TESTING=OFF:
https://cmake.org/cmake/help/latest/command/enable_testing.html

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


jrfonseca:
It sounds like `enable_testing` already honors `BUILD_TESTING` internally, so it shouldn't be necessary to wrap it in the if check, but no biggie.

Thanks.

Issue #699, opened 2020-12-29
Snapshots with frequenCy "frame" should always dump what is shown or about to be shown on the screen 


gerddie:
Currently, the contents of the active draw buffer is dumped, but this can also be a framebuffer attachment.  Some games, like *Slime Rancher* bind a fbo and clear it right before the buffer swap and clean it, so that the dumped image is all back regardless of what is actually shown on the screen.

PR #700, opened 2020-12-29
retrace: take end of frame shapshots always from back buffer

gerddie:
When specifying "--snapshot=frame" it can be assumed that the user
is interested in the renderung output that goes to the screen and
not to some framebuffer attachment that happens to be bound to be
bound to an active framebuffer and set to be the drawbuffer.

Consequently, override the frame capture logic in case the frame-end
call triggers the snapshot to get the backbuffer contents.

Closes: #699

### DIFF
--- a/retrace/d3dretrace.hpp
+++ b/retrace/d3dretrace.hpp
@@ -56,7 +56,7 @@ public:
     }
 
     image::Image *
-    getSnapshot(int n) {
+    getSnapshot(int n, bool backBuffer) {
         if ((n != 0) || !pLastDevice) {
             return NULL;
         }
--- a/retrace/glretrace_main.cpp
+++ b/retrace/glretrace_main.cpp
@@ -815,11 +815,11 @@ public:
     }
 
     image::Image *
-    getSnapshot(int n) override {
+    getSnapshot(int n, bool backBuffer) override {
         if (!glretrace::getCurrentContext()) {
             return NULL;
         }
-        return glstate::getDrawBufferImage(n);
+        return glstate::getDrawBufferImage(n, backBuffer);
     }
 
     bool
--- a/retrace/glstate.hpp
+++ b/retrace/glstate.hpp
@@ -68,7 +68,7 @@ int
 getDrawBufferImageCount(void);
 
 image::Image *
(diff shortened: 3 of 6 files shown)


jrfonseca:
Makes sense.

I'd prefer `s/endOfFrame/backBuffer/`, otherwise looks good.  (At some point it might make sense to have separate methods for rendertargets vs drawable, but we can tweak to our hears content later.)

You'll also need to adjust the D3D implementation of these interfaces -- it's fine to ignore this flag for now.  I'll try to make it easier to build apitrace with MinGW, by adding scripts to ci/... in the near future.

gerddie:
parameter renamed and d3d call changed (Appveyor check pending). The mingw build seems to have problems though.

gerddie:
Seems that this is now passing all tests.

jrfonseca:
> Seems that this is now passing all tests.

Looks good.  Feel free to commit.

> I can now dedicate some time to help with maintaining apitrace. First I will look out for issues and PRs that may be of interest to the use in the Mesa CI, so if there is something OpenGL specific you'd like some help with, just give me a pointer.

Sounds good.

I can't find an open issue, but one current limitation glretrace has which affects the usefulness of use for Mesa CI, is that glretrace doesn't use the same same visuals that were used while tracing.  The rationale is explained in https://github.com/apitrace/apitrace/blob/master/docs/BUGS.markdown#retracing but the end result is that, even if a GL app uses 4x MSAA backbuffer, glretrace will neve use MSAA.  There isn't any MSAA code coverage, except for MSAA FBOs.

Ideally the code on https://github.com/apitrace/apitrace/blob/master/retrace/glretrace_glx.cpp should keep track of the visual configs (such as sample count) and then pass it to https://github.com/apitrace/apitrace/blob/master/retrace/glws_glx.cpp when retracing.  Ditto for other windowing APIs (EGL, WGL, etc.)

> I'll also try to do some reviewing, but since I don't know many parts of the code base yet, this will probably not yet be to a point that no further reviewing is needed.

I have some high level overview of the source tree on https://github.com/apitrace/apitrace/blob/master/docs/HACKING.markdown

Thanks!

PR #702, opened 2021-01-19
Initialise application name

AnyOldName3:
`QSettings` requires that the application and organisation names are initialised before it can use an implicit settings location. When
QApiTrace uses `QSettings`, it doesn't set an explicit location, so the implicit one must be valid.

Without this, the *Recent Launches* list remains empty, at least on Windows. I've had it work on Ubuntu, though, so presumably something's setting the names automatically there.


jrfonseca:
I'm no Qt expert, but it looks sensible to me.

/cc @zackr

PR #705, opened 2021-01-29
Update bunded libbacktrace for dwarf-5 support

manisandro:
GCC 10 changed from defaulting from dwarf-v4 to dwarf-v5. The bundled libbacktrace is somewhat old and does not yet support dwarf-5. This results in libbacktrace_btest failing with

    unrecognized DWARF version in .debug_info at 6

Note: libbacktrace_btest will fail with current cmake versions if run through ctest due to [1], specifically it will fail with

    ERROR: descriptor 3 still open after tests complete

Even though this is a cmake bug, perhaps it might be appropriate to adapt btest.c to omit checking whether fd 3 is still open. I can modify the PR accordingly if desired.

[1] https://gitlab.kitware.com/cmake/cmake/-/issues/18863

### DIFF
--- a/thirdparty/libbacktrace/CMakeLists.txt
+++ b/thirdparty/libbacktrace/CMakeLists.txt
@@ -152,6 +152,7 @@ add_convenience_library (backtrace EXCLUDE_FROM_ALL
     print.c
     state.c
     sort.c
+    testlib.c
 )
 
 add_executable (btest btest.c)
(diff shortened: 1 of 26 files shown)


jrfonseca:
> Even though this is a cmake bug, perhaps it might be appropriate to adapt btest.c to omit checking whether fd 3 is still open. 

Sounds good.

> I can modify the PR accordingly if desired.

Please do.  Thanks.

manisandro:
Ok I've modified the PR

jrfonseca:
Thanks

PR #708, opened 2021-02-07
Makes MSVC version check earlier

vladislavbelov:
Prevents from configuring dependent libraries for a wrong/unsupported compiler.


jrfonseca:
Makes sense.  Thanks

PR #709, opened 2021-02-12
Gltrim: add handling various function calls and fix used of pipelines 

gerddie:
Add support for glClipControl. glCreateBuffers, glTextureBarrier,  glVertexAttrib(Binding|Format|Divisor), 
add all bound buffers when drawing as a dependency, and emit buffer that is written to by using memcopy in a preserved frame. These changes make it possible to trim traces obtained from Tomb Raider 2013.


jrfonseca:
I glanced through.  LGTM.

PR #710, opened 2021-02-19
Gltrim add copysubdata and indirect draw

gerddie:
gltrim: 

* Handle indirect and instanced draw calls
* Handle CopyBufferSubData and CopyNamedBufferSubData
* 
Some of these calls are used with "The Dark Mod"


jrfonseca:
LGTM

PR #713, opened 2021-02-25
gltrim: allow unordered call IDs when copying calls out of a trace

gerddie:
It seems that call IDs in traces are not necessarily ordered,
so don't rely on an ascending order. 

This fixes trimming a Tom Raider 2013 trace.

@jrfonseca This is a workaround, I guess the real fix would be to make sure that the call IDs are indeed ascending, but I didn't really find the code that is responsible for writing the calls in any order. If you can point me to it, I can try to figure out why the call IDs might be written out of order.

### DIFF
--- a/frametrim/ft_frametrimmer.cpp
+++ b/frametrim/ft_frametrimmer.cpp
@@ -72,6 +72,7 @@ struct FrameTrimmeImpl {
     void endTargetFrame();
 
     std::vector<unsigned> getSortedCallIds();
+    std::unordered_set<unsigned> getUniqueCallIds();
 
     static unsigned equalChars(const char *l, const char *r);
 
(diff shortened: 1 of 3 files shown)


jrfonseca:
This is by design.

When recording, calls are recorded asynchronously, that is, two calls can overlap.

Calls are numbered according the order control flow _entered_ the call.  But they are executed the order control flow _left_ the call (ie, the order calls become complete).   Therefore call numbers are not monotonically increasing.

We could change this by keeping a buffer of all unordered calls, but it means that if one thread was stuck, or took a lot of time, we might need to read the an enormous amount of calls into memory.

This has been brought up on other issues.  I can't find them.  I really should document this on the BUGS.markdown known issues.

gerddie:
As you can see in the patch, I can easily deal with this problem, so I don't really think it would be necessary to change this when tracing. Maybe it would be of interest to add another tool that can rewrite the trace ordering the calls.

jrfonseca:
The PR itself looks good to me.

PR #719, opened 2021-03-02
retrace: Add an option to force using the back buffer as snapshot source

gerddie:
Defaulting to reading from the back buffer when snapshots are taken at
frame frequency broke some test that only use FBOs and not always
clear the back buffer properly so that garbage was read back.
Add an option that the caller can force reading from the backbuffer,
but default to the old behaviour of reading back from the currently
bound draw buffer.

Fixes: 3d7d6e16606542e3255e03d8e99335aece7e5cdf
  retrace: take end of frame shapshots always from back buffer

### DIFF
--- a/retrace/retrace_main.cpp
+++ b/retrace/retrace_main.cpp
@@ -120,6 +120,7 @@ bool useCallNos = true;
 bool singleThread = false;
 bool ignoreRetvals = false;
 bool contextCheck = true;
+bool snapshotForceBackbuffer = false;
 int64_t minCpuTime = 1000;
 
 unsigned frameNo = 0;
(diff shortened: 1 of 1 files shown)


jrfonseca:
This is all subtle..

I know snapshots are tricky, especailly SwapBuffers as often leaves backbuffers corrupted, which is why we.

> Defaulting to reading from the back buffer when snapshots are taken at frame frequency broke some test that only use FBOs and not always clear the back buffer properly so that garbage was read back.

But I still don't understand what's going on: what kind of call exactly was this?  SwapBuffers, glFrameTerminatorGREMEDY, something else?

At any rate, given this reverts the default behavior to the previous known good behavior, I'm find with it as is.  We can always tweak things later.  Therefore LGTM.

gerddie:
The tests that were failing are actually part of the gltrim test suite, and stupid me didn't run them when I proposed the other PR. The failing tests  were taken from the GLES cts, the readback that they test is from an fbo, and the backbuffer is not cleared, so that there can be garbage.

PR #723, opened 2021-03-04
replay: Add option to execute and check queries

gerddie:
By default reading back query results is skipped, this commit adds the option to actually read back the query result and then either ignore it (useful for performance analysis, because reading back the result may trigger some synchronization, e.g. with the virgl driver), or check the result against the previously recorded one.

### DIFF
--- a/retrace/glretrace.py
+++ b/retrace/glretrace.py
@@ -175,15 +175,17 @@ class GlRetracer(Retracer):
             print(r'        return;')
             print(r'    }')
 
-        # When no query buffer object is bound, glGetQueryObject is a no-op.
+        # When no query buffer object is bound, and we don't request that glGetQueryObject
+        # is run than glGetQueryObject is a no-op.
         if function.name.startswith('glGetQueryObject'):
             print(r'    GLint _query_buffer = 0;')
             print(r'    if (currentContext && currentContext->features().query_buffer_object) {')
             print(r'        glGetIntegerv(GL_QUERY_BUFFER_BINDING, &_query_buffer);')
             print(r'    }')
-            print(r'    if (!_query_buffer) {')
+            print(r'    if (!_query_buffer && retrace::queryHandling == retrace::QUERY_SKIP) {')
             print(r'        return;')
             print(r'    }')
+            print(r'wait_for_query_result:')
 
         # Pre-snapshots
         if self.bind_framebuffer_function_regex.match(function.name):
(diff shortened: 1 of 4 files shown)


retrace/retrace.hpp
```diff
@@ -103,6 +103,14 @@ class ScopedAllocator : public ::ScopedAllocator
 };
 
 
+enum EQueryHadling {
```
jrfonseca, line 106: `EQueryHadling` -> `QueryHandling`

jrfonseca (commented):
Looks great, minus the typo. Thanks

gerddie:
There is still missing the part to compare the query result while replaying with the recorded one. I'm not sure where I have to change the code so that the visitor that created the loop iterating over the returned data actually checks the result. Any pointers?

gerddie:
The check is added now

jrfonseca:
> There is still missing the part to compare the query result while replaying with the recorded one. I'm not sure where I have to change the code so that the visitor that created the loop iterating over the returned data actually checks the result. Any pointers?

Rather than mucking with visitors, it might be easier to just print out the code gets the value.  For example

```c++
    const trace::Array *_params = call.arg(2).toArray();
    if (_params && _params->values.size() >= 1) {
         _result = _params->values[0]->toUInt();
    }
```

Of course, to `toUInt` might need to change to `toSInt` for some of the calls.

It would be nice if there as a method/function template that extracted the value of `trace::Value` to a variable automatically.  We could add such thing too.

jrfonseca:
Oops.  I just noticed you already fixed this.

You just looks great overall.

retrace/glretrace.py
```diff
@@ -196,6 +198,24 @@ def retraceFunctionBody(self, function):
...
+        if function.name.startswith('glGetQueryObject'):
+           print(r'    if (_query_buffer || retrace::queryHandling != retrace::QUERY_SKIP) {')
+           print(r'        auto query_result = call.arg(2).toArray();')
+           print(r'        assert(query_result && query_result->values.size() == 1);')
+           print(r'        auto expect = static_cast<decltype(retval)>(query_result->values[0]->toUInt());')
+           print(r'        if (call.arg(1).toUInt() == GL_QUERY_RESULT_AVAILABLE) {')
+           print(r'            if (expect == 1 && retval == 0)')
+           print(r'                goto wait_for_query_result;')
+           print(r'        } else if (!_query_buffer && expect != retval) {')
+           print(r'           std::cout << "Warning: query returned " << retval << " but trace contained " << expect << "\n";')
```
jrfonseca, line 215: let's use `retrace::warning` so this gets plugged into the UI correctly.

retrace/glretrace.py
```diff
@@ -196,6 +198,24 @@ def retraceFunctionBody(self, function):
 
         Retracer.retraceFunctionBody(self, function)
 
+        # If we have a query buffer or if we execute the query we have to loop until the
+        # query result is actually available to get the proper result - for the following
+        # execution if the query buffer is used or for the check to make sense, and if we
+        # just want to execute the query for timing purpouses we also should wait
+        # for the result.
+        if function.name.startswith('glGetQueryObject'):
+           print(r'    if (_query_buffer || retrace::queryHandling != retrace::QUERY_SKIP) {')
```
jrfonseca, line 207: I think this should be `!_query_buffer` no?

When there's a query buffer bound, the value is not written to `params`, but rather written into the query buffer, into the offset given by `params`.  See https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetQueryObject.xhtml

jrfonseca (commented):
Otherwise LGTM

retrace/glretrace.py
jrfonseca: In fact, I think this shoud be 

```c++
    if (!_query_buffer && retrace::queryHandling != retrace::QUERY_SKIP) {
```

gerddie: I know, I just had a laps in logic, not thinking  that for checking whether the result available it is follows the same rule. But there was also an error when the query result call was run but the result not supposed to be checked. I have updated the MR accordingly.

retrace/glretrace.py
gerddie: Changed.

jrfonseca:
Update looks great.  Thanks.

Issue #715, opened 2021-03-02
[Feature] Compare return value recorded in trace with current machine


AnyOldName3:
It's possible that this feature exists already and I just don't know how to use it. If so, I'd appreciate being pointed in the right direction.

It's my understanding that the return values and out parameters displayed in QAPITrace are from the machine the trace was recorded on, not the one it's being replayed on. I'm currently trying to debug an issue with occlusion queries, where one certain driver and hardware combination gives no samples passing, and every other combination gives thousands. Specifically, I see `glGetQueryObjectiv(id = 1, pname = GL_QUERY_RESULT, params = [44887])` and `glGetQueryObjectiv(id = 1, pname = GL_QUERY_RESULT, params = [0])` depending on the machine the trace is captured on.

What I'd like to be able to do is see what the same call would return on the machine the trace is being replayed on - if I could, then if it matched, I'd know something was different about the renderer state on the two machines, and if it didn't match, I'd know the two OpenGL implementations thought a different number of samples were passing despite the state being identical.

jrfonseca:
This feature doesn't exists, but I agree it would be useful.

It wouldn't be too complicated: one would need to modify https://github.com/apitrace/apitrace/blob/a22dda1ac2f27cd014ac7a16e7b7b6ebc9f14ae1/retrace/glretrace.py#L178-L187 to generate the code to still execute `glGetQueryObject*` calls when no query buffer is bound (instead of skippnig), and compare the result with the one recorded, and emit a warning when they are different.

However I don't think I'll have time to look into it soon,  as I have a never ending backlog of issues and PRs to sift through..

gerddie:
I'll take a look into it. We were already discussing that it would be nice if queries would be executed, because we want to debug reading back values in virglrenderer

gerddie:
Would it make sense to make this an optional feature (as per command line parameter)?

jrfonseca:
> I'll take a look into it. We were already discussing that it would be nice if queries would be executed, because we want to debug reading back values in virglrenderer

Thanks!

> Would it make sense to make this an optional feature (as per command line parameter)?

Yes, I think a command line parameter (e.g, `--check`) makes a lot of sense.  Not only some queries are inconsistent (e.g, time queries), but some like pipeline stats are very specific to GPU architectures.

AnyOldName3:
[...]

PR #724, opened 2021-03-16
retrace: Add option to pass a query result check threshold

gerddie:
In order to reduce the noise when comparing recorded query values
to evaluated ones while retracing add an option to be able to
specify a threshold for the result difference.
With a threshold > 0 " --query-handling check" will be enabled too.

Related: #715

### DIFF
--- a/gui/mainwindow.cpp
+++ b/gui/mainwindow.cpp
@@ -250,6 +250,16 @@ void MainWindow::replayStart()
     dlgUi.coreProfileCB->setChecked(
         m_retracer->isCoreProfile());
 
+    dlgUi.queryHandlingSelector->addItem("Skip");
+    dlgUi.queryHandlingSelector->addItem("Run");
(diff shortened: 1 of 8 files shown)


jrfonseca:
Some builds are failing with `error: no member named 'setPlaceholderText' in 'QComboBox'`  Either this method is too old or too new?

jrfonseca:
Oh, it's not your fault!  It seems that GitHub Actions migrated to 20.04 and this triggered troubles.  I'm looking into it...

jrfonseca:
I've fixed the master builds, but it turned out to be something completely different.  The `setPlaceholderText` seems introduced by this change.

gerddie:
> I've fixed the master builds, but it turned out to be something completely different. The `setPlaceholderText` seems introduced by this change.

I've removed it and set the initial text instead, It seems to work nicely.-

jrfonseca:
Thanks.

Issue #690, opened 2020-11-10
MSAA disabled when replaying a trace


AnyOldName3:
Tested on Windows with OpenMW (https://gitlab.com/OpenMW/openmw) which uses OpenGL 2.1 and renders to the main window's framebuffer.

When capturing a trace, MSAA works correctly, but when replaying it, it doesn't. It *looks* to me like the WGL calls at the start of the trace are requesting the pixel format number to be one with MSAA, but the pixel format number returned and used is one without. I'm not sure whether the WGL calls in the trace are what OpenMW calls directly or whether they've been adjusted by APITrace as we use SDL2 to abstract this away.

jrfonseca:
Yes, this is a know limitation of glretrace.  It would be fine to fix it eventually, but it implies a fair amount of work, due to the fact that WGL/GLX/CGL calls are never replayed verbatim, but go through an emulation layer to allow replaying on one OS a trace captured on a different OS.

jrfonseca:
@gerddie  , FYI, this is the issue I mentioned on https://github.com/apitrace/apitrace/pull/700#issuecomment-773411284

PR #725, opened 2021-03-16
glretrace: use samples value recorded in trace to set MSAA samples

gerddie:
Check the calls that configure the framebuffer format and use the
requested samples value unless this value was set from the command
line.

WIP: currently calling the framebuffer config twice with different samples values will result in ignoring the second value if the first one was != 1. 

Related: #690 

@jrfonseca could you skim over it to see whether this makes any sense?

### DIFF
--- a/helpers/glfeatures.hpp
+++ b/helpers/glfeatures.hpp
@@ -49,6 +49,7 @@ enum Api {
 struct Profile {
     unsigned major:8;
     unsigned minor:8;
+    unsigned samples:8;
     unsigned api:1;
     unsigned core:1;
     unsigned forwardCompatible:1;
@@ -64,6 +65,7 @@ struct Profile {
         api = _api;
         major = _major;
         minor = _minor;
+        samples = 1;
         core = _core;
         forwardCompatible = _forwardCompatible;
     }
--- a/retrace/glretrace.hpp
+++ b/retrace/glretrace.hpp
@@ -132,6 +132,9 @@ getCurrentContext(void) {
 int
 parseAttrib(const trace::Value *attribs, int param, int default_ = 0, int terminator = 0);
 
+void
+setSamples(trace::Call& call, int samples);
+
 glfeatures::Profile
 parseContextAttribList(const trace::Value *attribs);
 
--- a/retrace/glretrace_cgl.cpp
+++ b/retrace/glretrace_cgl.cpp
@@ -90,11 +90,14 @@ using namespace glretrace;
 
 typedef std::map<unsigned long long, glws::Drawable *> DrawableMap;
 typedef std::map<unsigned long long, Context *> ContextMap;
+typedef std::map<unsigned long long, int> FBConfigMap;
 
 // sid -> Drawable* map
 static DrawableMap drawable_map;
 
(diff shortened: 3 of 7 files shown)


jrfonseca:
> @jrfonseca could you skim over it to see whether this makes any sense?

@gerddie, yeah, looks the right direction.

Ideally we'd book-keep visual information like `samples` and `doubleBuffer` and use it on a per-drawable basis, instead of `retrace::` globals but this change is definitely a step step forward.  We can tackle that seperately.

Subsequently, it would be good to have some tests too.  I really need to find the time to integrate https://github.com/apitrace/apitrace-tests into apitrace/tests, so the tests are easier to run.

gerddie:
I've updated the MR to store the sampled on a per FBConfig basis and query the value when creating the context. It seems to work for glretrace, with eglretrace I get the right samples value, but the output doesn't look like MSAA is applied, for wgl I couldn't run any tests.

jrfonseca:
> I've updated the MR to store the sampled on a per FBConfig basis and query the value when creating the context. It seems to work for glretrace, with eglretrace I get the right samples value, but the output doesn't look like MSAA is applied, for wgl I couldn't run any tests.

Looks great.  

glXChooseFBConfig is not sufficient.  Take for example GLFW3, it doesn't use .  Instead it uses glXGetFBConfigs followed by glXGetFBConfigAttrib, and then chooses manually the config.  So for this to work, glretrace should also interpret calls like

    glXGetFBConfigAttrib(dpy = ..., config = foo, attribute = GLX_SAMPLES, value = &4) 

and use this information to populate `fbconfig_map[foo] = 4`.

One can modify https://github.com/apitrace/apitrace-tests/blob/master/apps/gl/tri.cpp to use MSAA to test this.  In fact, I think it would be a good idea on its own right to have a MSAA test.

That said, don't feel you need to nail every corner case before commit -- as long as we're doing forward progress and cause no regressions, partial improvements are great IMO.

gerddie:
I've added the check for glXGetFBConfigAttrib and fixed an issue when no sample count was given.

jrfonseca:
> I've added the check for glXGetFBConfigAttrib and fixed an issue when no sample count was given.

Looks great.  Thanks!

Regarding your outstanding WIP item:
> WIP: currently calling the framebuffer config twice with different samples values will result in ignoring the second value if the first one was != 1.

Can you elaborate the issue? 

I think that even without it this seems a step forward.  As long as it doesn't cause obvious errors crashes with existing traces I'm OK with commiting this.  And we can deal with any issues detected with existing traces after commiting.

gerddie:
I think the thing with the framebuffer config called twice with different sample values is not relevant, i.e. if I understand it correctly, different samples values will also result in different framebuffer configs. 

There was another issue I noted when I tried to implement the tests: sometimes glXMakeCurrent also creates the drawable,and it was using the default profile, therefore all the work with passing the samples value around was done in vain. I've added another commit (the first commit of the PR) to use the context profile  if available when getDrawable creates a new drawable. This fixed up things nicely. 

I've added a test: https://github.com/apitrace/apitrace-tests/pull/8

jrfonseca:
Thanks!

PR #726, opened 2021-03-17
Increase precision of image range spinners

AnyOldName3:
I needed to be able to restrict the range to 0.9998-1.0 to be able to see something far away in a depth buffer, so made this change locally as three decimal digits didn't let me specify such a small range. Presumably, someone else might want to visualise tiny depth buffer differences, too.


jrfonseca:
Sounds good.

PR #730, opened 2021-04-24
Add some instructions for bundling dependencies

dmbaturin:
Building on Fedora requires quite a few bundled dependencies, since the distro repos don't have them.

.gitmodules is quite easy to overlook, so I've added instructions to INSTALL.markdown


jrfonseca:
Thanks.  This applies to non-Linux too.  I'll move the paragraph afterwards.

Issue #729, opened 2021-04-21
Assertion or crash when using -loop with Brotli traces


tomeuv:
Currently, when using the --loop argument with a trace in Brotli format we hit the following assertion in debug builds:

```
../lib/trace/trace_file.cpp:58: virtual void trace::File::setCurrentOffset(const trace::File::Offset&): Assertion `0' failed.
Thread 1 "eglretrace" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x0000fffff79a5ea0 in __GI_abort () at abort.c:79
#2  0x0000fffff79b2878 in __assert_fail_base (fmt=0xfffff7ab0298 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0xaaaaab1267e0 "0", file=file@entry=0xaaaaab126768 "../lib/trace/trace_file.cpp", line=line@entry=58, 
    function=function@entry=0xaaaaab126798 "virtual void trace::File::setCurrentOffset(const trace::File::Offset&)") at assert.c:92
#3  0x0000fffff79b28dc in __GI___assert_fail (assertion=0xaaaaab1267e0 "0", file=0xaaaaab126768 "../lib/trace/trace_file.cpp", line=58, 
    function=0xaaaaab126798 "virtual void trace::File::setCurrentOffset(const trace::File::Offset&)") at assert.c:101
#4  0x0000aaaaaae54438 in trace::File::setCurrentOffset (this=0xaaaaab320c20, offset=...) at ../lib/trace/trace_file.cpp:58
#5  0x0000aaaaaae3c64c in trace::Parser::setBookmark (this=0xaaaaab320b20, bookmark=...) at ../lib/trace/trace_parser.cpp:173
#6  0x0000aaaaaae4eac0 in trace::LastFrameLoopParser::parse_call (this=0xaaaaab2ac750) at ../lib/trace/trace_parser_loop.cpp:93
#7  0x0000aaaaaaba3ac8 in retrace::RelayRunner::runLeg (this=0xaaaaab31dcc0, call=0xaaaaacf92760) at ../retrace/retrace_main.cpp:437
[...]

PR #734, opened 2021-05-05
retrace: Get --loop to work without rewinding

tomeuv:
This way we can benefit from the higher compression rates of Brotli when
doing performance tracking with --loop.

Signed-off-by: Tomeu Vizoso <tomeu.vizoso@collabora.com>

### DIFF
--- a/lib/trace/trace_parser_loop.cpp
+++ b/lib/trace/trace_parser_loop.cpp
@@ -36,6 +36,7 @@ public:
     LastFrameLoopParser(AbstractParser *p, int c) {
         parser = p;
         loopCount = c;
+        starts_new_frame = true;
     }
 
     ~LastFrameLoopParser() {
@@ -56,8 +57,8 @@ public:
     const Properties & getProperties(void) const override { return parser->getProperties(); }
 private:
     int loopCount;
+    bool starts_new_frame;
     AbstractParser *parser;
-    ParseBookmark lastFrameBegin, lastFrameEnd;
     std::vector<Call *> lastFrameCalls;
     std::vector<Call *>::iterator lastFrameIterator;
 };
@@ -66,16 +67,7 @@ private:
 bool
 LastFrameLoopParser::open(const char *filename)
 {
-    bool ret = parser->open(filename);
-    if (ret) {
-        /* If the user wants to loop we need to get a bookmark target. We
-         * usually get this after replaying a call that ends a frame, but
-         * for a trace that has only one frame we need to get it at the
-         * beginning. */
-        parser->getBookmark(lastFrameBegin);
-        lastFrameEnd = lastFrameBegin;
-    }
-    return ret;
+    return parser->open(filename);
 }
 
 Call *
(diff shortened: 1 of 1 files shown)


lib/trace/trace_parser_loop.cpp
```diff
@@ -87,34 +79,23 @@ LastFrameLoopParser::parse_call(void)
...
+        lastFrameCalls.push_back(call);
+        starts_new_frame = call->flags & trace::CALL_FLAG_END_FRAME;
+
+        lastFrameIterator = lastFrameCalls.begin();
+
         return call;
     }
 
     /* else */
     if (loopCount) {
```
gerddie, line 98: Given that now we have all the calls in the last frame available here, we might want to add a filtering step that removes all the calls that create or delete objects and are not paired. Creating more objects might crash the driver if the loop count is high. and destroying something might leave us with objects missing when the frame is replayed. But maybe that's for another PR.

lib/trace/trace_parser_loop.cpp
```diff
@@ -87,34 +79,23 @@ LastFrameLoopParser::parse_call(void)
 
     /* Restart last frame when looping is requested. */
     if (call) {
-        if (call->flags & trace::CALL_FLAG_END_FRAME) {
-            lastFrameBegin = lastFrameEnd;
-            parser->getBookmark(lastFrameEnd);
+        if (starts_new_frame) {
```
gerddie, line 82: We probably only want to store the calls only `if (loopCount)`

gerddie: With that changed this is `Reviewed-by: Gert Wollny <gert.wollny@collabora.com>`

lib/trace/trace_parser_loop.cpp
```diff
@@ -87,34 +79,23 @@ LastFrameLoopParser::parse_call(void)
...
-                if (current.next_call_no == lastFrameEnd.next_call_no) {
-                    /* end of frame */
-                    lastFrameIterator = lastFrameCalls.begin();
-                    break;
-                }
-            }
-        }
         call = *lastFrameIterator;
         ++lastFrameIterator;
         if (lastFrameIterator == lastFrameCalls.end()) {
```
gerddie, line 101: I guess this  deserves a `Closes: #729` in the commit message, right?

lib/trace/trace_parser_loop.cpp
tomeuv: IIUIC, loopCount will be zero only when we reach the last iteration, and by then this code doesn't execute.

lib/trace/trace_parser_loop.cpp
tomeuv: Good idea!

lib/trace/trace_parser_loop.cpp
gerddie: My bad,  I thought that this code is always run, but now I see that it is only run when looping is actually requested from the command line, sorry for the noise.

lib/trace/trace_parser_loop.cpp
gerddie: The *not destroying* part is actually a bit more difficult, because at this point the destroy commands are already run before entering the loop.

lib/trace/trace_parser_loop.cpp
tomeuv: Indeed, thanks!

lib/trace/trace_parser_loop.cpp
tomeuv: Yep, I also got to think that at first.

lib/trace/trace_parser_loop.cpp
tomeuv: Will keep this in mind as we do more testing with this, thanks!

PR #735, opened 2021-05-07
Various gltrim fixes

gerddie:
- keep MSAA info in trace (fixes the gltrim tests that broke when we started to replay with the MSAA info)
- fix buffer bindings, mapping and unmapping 
- fix reading textures from unpack buffers 
- add proper support for BindVertexArray


jrfonseca:
I glanced through them.  They look alright to me.

PR #737, opened 2021-05-12
retrace: Handle timestamp counter roll-overs

tomeuv:
Ask the OpenGL implementation what resolution it supports and handle any
roll overs that could have happened.

Signed-off-by: Tomeu Vizoso <tomeu.vizoso@collabora.com>

### DIFF
--- a/retrace/metric_backend_opengl.cpp
+++ b/retrace/metric_backend_opengl.cpp
@@ -23,6 +23,8 @@
  *
  **************************************************************************/
 
+#include <math.h>
+
 #include "metric_backend_opengl.hpp"
 #include "os_time.hpp"
 #include "os_memory.hpp"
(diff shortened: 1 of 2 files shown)


retrace/metric_backend_opengl.cpp
```diff
@@ -307,15 +311,20 @@ void MetricBackend_opengl::beginPass() {
...
         while (!queries[i].empty()) {
             auto &query = queries[i].front();
             if (metrics[METRIC_GPU_START].profiled[i]) {
                 glGetQueryObjecti64v(query[QUERY_GPU_START], GL_QUERY_RESULT,
-                                     &gpuStart);
-                int64_t value = gpuStart - baseTime;
+                                      &timestamp);
+                if (timestamp < gpuStart) {
+                    // Correct clock domain of baseTime after roll-over
+                    baseTime -= pow(2, timestamp_bits_precision);
```
jrfonseca, line 324: These are all integer quantities, whereas `pow` is a function of doubles.

Let's use `UINT64_C(2) << timestamp_bits_precision` instead, here and below.

retrace/metric_backend_opengl.cpp
```diff
@@ -83,6 +85,8 @@ MetricBackend_opengl::MetricBackend_opengl(glretrace::Context* context,
     supportsElapsed     = context->hasExtension("GL_EXT_timer_query") || supportsTimestamp;
     supportsOcclusion   = currentProfile.versionGreaterOrEqual(glfeatures::API_GL, 1, 5);
 
+    glGetQueryiv(GL_TIMESTAMP, GL_QUERY_COUNTER_BITS, &timestamp_bits_precision);
```
jrfonseca, line 88: We should add here

```c++
    if (timestamp_bits_precision >= 64) {
         timestamp_bits_precision = 0;
    }
```
Because we our counters are also 64 bits.

jrfonseca (commented):
Otherwise LGTM.

jrfonseca:
Thanks.

PR #738, opened 2021-05-13
Initial cleanup/preparations for D3D12 tracing

Joshua-Ashton:
The initial cleanups so I can start upstreaming this.

Doesn't contain any D3D12-specific stuff yet, just want to get simple cleanup and base changes out of the way first before we deal with that as the commits get much heavier there and will probably come as 1-3 per PR to ease review.

### DIFF
--- a/lib/os/os_string.hpp
+++ b/lib/os/os_string.hpp
@@ -228,7 +228,7 @@ public:
         va_end(args_copy);
 
         assert(length >= 0);
-        size_t size = length + 1;
+        size_t size = size_t(length) + 1;
 
         String path(size);
 
--- a/lib/trace/trace_dump.cpp
+++ b/lib/trace/trace_dump.cpp
@@ -136,11 +136,9 @@ Dumper::visitMembers(Struct *s, const char *sep) {
         if (!memberName || !*memberName) {
             // Anonymous structure
             Struct *memberStruct = memberValue->toStruct();
-            assert(memberStruct);
-            if (memberStruct) {
+            if (memberStruct)
                 sep = visitMembers(memberStruct, sep);
-                continue;
-            }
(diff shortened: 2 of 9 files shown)


dave-juicelabs (approved):
LGTM

jrfonseca:
@Joshua-Ashton, thanks for doing this.  Appologies for the delay reviewing.

All commits look great and uncontroversial,  except one.

The commit I'm unsure is https://github.com/apitrace/apitrace/pull/738/commits/1cd9ea321043e9bcab9d0e2321bd3cf8fd5c5f02

The way I've tackled this issue until now is to describe this as malloc+memcpy+free.

For example, if there's a blob whose memory pointer address is important, instead of

    OpenExistingHeapFromAddress(blob(0x12345678, 12345), ...)

we would have

    0x12345678 = malloc(12345)
    memcpy(0x12345678, blob(12345), 1234)
    OpenExistingHeapFromAddress(0x12345678, ...)

where 0x12345678/12345 matches the VirtualAlloc arguments (or more accurately, the information returned by VirtualQuery.)
possibly followed by

    free(0x12345678)

when apitrace detected that that addressed was VirtualFree'd.

ApiTrace already has code to ensure the 0x12345678 address (and any address inside the malloced range) is swizlled properly.

I believe this is more versatile, and although I'm not familiar with ID3D12Device3::OpenExistingHeapFromAddress it seems to fit the bill.

Would that work?

Joshua-Ashton:
I think it would, thanks for the heads up. I will test that out.

Joshua-Ashton:
I have dropped the blob commit, hopefully this should be good now ^^

jrfonseca:
Thanks.

PR #739, opened 2021-05-17
gltrim: some more fixes and handling of additional calls

gerddie:
- glRenderMode,
- glPushAttrib, glPopAttrib, glClientPushAttrib, glPopClientAttrib
- glNamedProgramLocalParameter
- glNamedProgramEnvParameter
- glActiveStencilFace
- glMultiTexCoord
- OpenGL 2.1 allows a texture to be created with the bind call so add the texture object on bind if it doesn't exist
- avoid segfault when the vertex array is unbound


jrfonseca:
LGTM.

PR #741, opened 2021-05-20
specs/glxapi: Make GLX_SAMPELES and GLX_SAMPLE_BUFFERS accept int

gerddie:
The call to glXChooseFBConfig defines the attrib_list as pointer to int.
Fixes replaying and traces of "Lifeless Planet"

Remark: Maybe all the values should be defined as Int.

### DIFF
--- a/specs/glxapi.py
+++ b/specs/glxapi.py
@@ -298,8 +298,8 @@ GLXCommonAttribs = [
     ('GLX_TRANSPARENT_GREEN_VALUE', Int),
     ('GLX_TRANSPARENT_BLUE_VALUE', Int),
     ('GLX_TRANSPARENT_ALPHA_VALUE', Int),
-    ('GLX_SAMPLE_BUFFERS', UInt),
-    ('GLX_SAMPLES', UInt),
+    ('GLX_SAMPLE_BUFFERS', Int),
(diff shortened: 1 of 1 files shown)


jrfonseca:
> Maybe all the values should be defined as Int.

Indeed..

But I'm worried about values such as `GLX_DONT_CARE`, which is defined as `0xFFFFFFFF`.  It would be nice if the traces still recognized `GLX_DONT_CARE` as `GLX_DONT_CARE`, and not `-1` or `0xFFFFFFFF`.

> Fixes replaying and traces of "Lifeless Planet"

Could you elaborate what was the problem exactly?

jrfonseca:
Never mind, `GLXEnum` is defined as `FakeEnum(Int, [...])` so it's using integers already.

Still I'd like to understand how/why `UInt` vs `Int` made a difference.

jrfonseca:
I see, because of the `assert(static_cast<signed long long>(value) >= 0);` in `UInt::toSInt`..

We can go with this, but I get the feeling this might pop up elsewhere and `UInt::toSInt` should become more lenient.

jrfonseca:
Sorry for the delay BTW.  Outlook 265 decided to send mant Github / Freedesktop notifications to spam... I've added the email address to exclusion lists now.

PR #742, opened 2021-05-22
cli: fix dump-images typos.

tanty:
Fixes: 4edd6294 ("glretrace, dump-images: Accept --call-nos=no to get snapshots with call numbers")

Signed-off-by: Andres Gomez <tanty@igalia.com>


jrfonseca:
Thanks

Issue #740, opened 2021-05-18
Cannot replay MacOS trace on Windows due to context creation failure


AnyOldName3:
I've got a trace from a user with MacOS that I can't replay on Windows. I think this is supposed to be supported as I've never had trouble running Windows or Linux traces on a different machine to where they were recorded. When calling `glretrace` via the command line, it prints:
```
1 @0 CGLRetainPixelFormat(pix = 0x7fc403f765c0) = 0x7fc403f765c0
1: warning: unsupported CGLRetainPixelFormat call
2 @0 CGLReleasePixelFormat(pix = 0x7fc403f765c0)
2: warning: unsupported CGLReleasePixelFormat call
4 @0 CGLRetainContext(ctx = 0x7fc40586f200) = 0x7fc40586f200
4: warning: unsupported CGLRetainContext call
5 @0 CGLReleaseContext(ctx = 0x7fc40586f200)
5: warning: unsupported CGLReleaseContext call
6 @0 CGLReleasePixelFormat(pix = 0x7fc403f765c0)
6: warning: unsupported CGLReleasePixelFormat call
error: failed to make current OpenGL context and drawable
```

The first frame of the trace, which is enough to reproduce the issue is attached:

[buggy aca5d2fc1.418.trim.trace.zip](https://github.com/apitrace/apitrace/files/6504874/buggy.aca5d2fc1.418.trim.trace.zip)

Issue #662, opened 2020-04-30
macOS strikes back


Lord-Kamina:
First; I don't know if anything came out of #594  in the end but I can build and trace just fine now.
However, opening my trace and trying to profile or doing mostly anything crashes on an assertion:

```
Process:               glretrace [36794]
Path:                  /Volumes/VOLUME/*/glretrace
Identifier:            glretrace
Version:               0
Code Type:             X86-64 (Native)
Parent Process:        qapitrace [34423]
Responsible:           glretrace [36794]
User ID:               501

Date/Time:             2020-04-29 20:26:11.107 -0400
OS Version:            Mac OS X 10.14.6 (18G3020)
Report Version:        12
Bridge OS Version:     3.0 (14Y908)
Anonymous UUID:        EA9784FA-1965-044C-73CA-79464B4271C5


Time Awake Since Boot: 4900 seconds

System Integrity Protection: disabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
Assertion failed: (0), function toUIntPtr, file /Users/Koji/apitrace/lib/trace/trace_model.cpp, line 235.
 

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff7ada32c2 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff7ae5ebf1 pthread_kill + 284
2   libsystem_c.dylib             	0x00007fff7ad0d6a6 abort + 127
3   libsystem_c.dylib             	0x00007fff7acd620d __assert_rtn + 324
4   glretrace                     	0x00000001069ec09b trace::Value::toUIntPtr() const + 43
5   glretrace                     	0x00000001068ad008 retrace_CGLGetPBuffer(trace::Call&) + 104
6   glretrace                     	0x00000001066f3f4c retrace::Retracer::retrace(trace::Call&) + 588
7   glretrace                     	0x00000001066fdb29 retrace::retraceCall(trace::Call*) + 41
8   glretrace                     	0x00000001066fd971 retrace::RelayRunner::runLeg(trace::Call*) + 161
9   glretrace                     	0x00000001066f75aa retrace::RelayRunner::runRace() + 266
[...]

PR #743, opened 2021-06-08
fix: replay MacOS trace on Windows

remoe:
This should fix https://github.com/apitrace/apitrace/issues/740 and #662 

It doesn't fix "error: failed to make current OpenGL context and drawable". This is another issue.


remoe:
I can play the trace on Mac when I remove this check ( ! ):

https://github.com/apitrace/apitrace/blob/master/retrace/glretrace_ws.cpp#L188

Sometimes SetCurrentContext failed ?!

jrfonseca:
Sorry for the delay.  I've merged this after a few tweaks.  Indeed the problem was that pbuffer argument here is not a straight pbuffer, but rather a pointer to a pbuffer.

PR #745, opened 2021-06-12
Fix dumping vertex attributes

10110111:
When `AttribDesc::AttribDesc` fills in the data fields, it tries to find number of columns from size and type only. This works OK for uniforms (and presumably feedback transform attributes), because they can have type like `GL_FLOAT_VEC3`, but this fails for vertex attributes, where types can only be scalar (those acceptable to `glVertexAttribPointer`).

This patch special-cases filling in the data fields for vertex attributes, which then lets `qapitrace` show the correct structure of vertex buffers.

### DIFF
--- a/retrace/glstate_shaders.cpp
+++ b/retrace/glstate_shaders.cpp
@@ -283,6 +283,26 @@ struct AttribDesc
     }
 };
 
+struct VertexAttribDesc : AttribDesc
+{
+    VertexAttribDesc()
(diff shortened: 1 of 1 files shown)


jrfonseca:
Makes sense.

The new `VertexAttribDesc` structure type could be avoided, but it's not a big deal.

jrfonseca:
Thanks

PR #746, opened 2021-06-14
gltrim: Handle glNormalPointer, glMultiTexCoord and a few  glX calls

gerddie:
This is required for trimming "Sir, You Are Being Hunted" (but #741 is also required)


jrfonseca:
LGTM

PR #748, opened 2021-06-21
workflows: Update package list on ubuntu64*

gerddie:
Apparently Ubuntu may also drop package versions in LTS
distros, so always run "apt-get update" before trying to
install packages.

case in point: https://github.com/gerddie/apitrace/runs/2873510746?check_suite_focus=true


jrfonseca:
Makes sense.  I thought that upstream was regularly doing update.  Thanks

PR #749, opened 2021-07-02
Add support for newer DXGI, misc fixes

Joshua-Ashton:
- Adds support for IDXGIFactory7
 - Fixes non-default adapters in `EnumAdapterByGpuPreference`
 - Handles returning non-default adapters for EnumAdapterByLuid (these are never going to match up)
 - Fix the `intersects` region call on 64-bit

### DIFF
--- a/specs/dxgi.py
+++ b/specs/dxgi.py
@@ -1136,6 +1136,7 @@ DXGI_GPU_PREFERENCE = Enum('DXGI_GPU_PREFERENCE', [
 ])
 
 IDXGIFactory6 = Interface('IDXGIFactory6', IDXGIFactory5)
+IDXGIFactory7 = Interface('IDXGIFactory7', IDXGIFactory6)
 IDXGIAdapter4 = Interface('IDXGIAdapter4', IDXGIAdapter3)
 IDXGIOutput6 = Interface('IDXGIOutput6', IDXGIOutput5)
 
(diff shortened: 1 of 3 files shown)


specs/dxgi.py
```diff
@@ -1149,7 +1150,13 @@
     StdMethod(HRESULT, 'EnumAdapterByGpuPreference', [(UINT, 'Adapter'), (DXGI_GPU_PREFERENCE, 'GpuPreference'), (REFIID, 'riid'), Out(Pointer(ObjPointer(Void)), 'ppvAdapter')]),
 ]
 
+IDXGIFactory7.methods += [
+    StdMethod(HRESULT, 'RegisterAdaptersChangedEvent', [(HANDLE, 'hEvent'), (Pointer(DWORD), 'pdwCookie')], sideeffects=False),
```
jrfonseca, line 1154: Should be `Out(Pointer(DWORD), 'pdwCookie')`, to tell it's an output parameter.

jrfonseca (changes requested):
Otherwise looks great/

specs/dxgi.py
Joshua-Ashton: I went on a vacation shortly after this and forgot, oops! Will fix tomorrow.

Joshua-Ashton: Fixed!

PR #750, opened 2021-07-06
Revert "os: Emulate C++11 threads on MinGW."

cwabbott0:
We cannot rely on win32 threads because (1) other projects like dxvk require winpthreads and (2) some distros (like Fedora) don't even provide mingw with win32 threads. Revert this to fix the build breakage without forcing users to build their own mingw.


jrfonseca:
I don't understand (1): what dxvk requires or not has nothing to do with apitrace, since these projects are built separately, and it should be possible to use apitrace built with win32 threads with

(2) is fair enough though.

That said, I rather continue building apitrace for mingw with win32 threads.  Therefore instead of reverting, I can use pthreads automatically.  I can post a fix.

I don't see why reverting https://github.com/apitrace/apitrace/commit/f3306538a7813ed4320a49c4833443828ef45365 is necessary.

cwabbott0:
wrt (1), I was under the impression that most ways of distributing mingw at most allow you to pick which one, not install both versions side-by-side. Of course they're not related, but what happens when you need to build both?

I tried using pthreads initially, i.e. just commenting out the error, but got a build error because (at least for me) `<threads>` on my system uses `__out` as a variable name and it's included after some windows file which defines it as a macro (I think). Not your fault, but it seems getting winpthreads to actually work isn't trivial.

I needed to revert that commit because without it the os threads test fails to compile. You can see the CI also failed.

jrfonseca:
> Of course they're not related, but what happens when you need to build both?

I see now what you mean.  Note for example Debian bundles both win32 and pthreads versions, and it's trivial to pick one over the other: either use different gcc prefix, or pass `-posix`.

Regarding `__out`, there's a different fix for that: one just needs to ensure `#include <threads>` is added to `compat/winsdk_compat.h` to ensure the header is included _before_ `__out` macros are defined.

This is quite annoying.  Another solution could be to use C-PP scopes and undefine things like `__out` before including such headers...

PR #753, opened 2021-07-12
glretrace: Don't try to get pointer for buffer '0' in DeleteBuffers

gerddie:
glDeleteBuffers ignores non-existing buffers names, and especially '0',
but calling glGetNamedBufferPointerv on a non-existing buffer results in
an error 'GL_INVALID_OPERATION'. To avoid this error skip over buffer '0'
when handling the region pointers.


jrfonseca:
LGTM

PR #754, opened 2021-07-12
GLtrim: track shared and per-context resources 

gerddie:
Blender 3D 2.80+ uses shared contexts,with this series its traces can be trimmed.


jrfonseca:
LGTM

PR #758, opened 2021-08-03
glretrace: Handle glViewport better.

LeptonWu:
If a user call glViewport with non default framebuffer bind first,
and then bind the default framebuffer to draw, the current apitrace
won't show anything on screen and complain glViewport hasn't been
called yet. Fix this by caching the width/height value in current
context.

### DIFF
--- a/retrace/glretrace.hpp
+++ b/retrace/glretrace.hpp
@@ -80,6 +80,8 @@ public:
     bool insideBeginEnd = false;
     bool insideList = false;
     bool needsFlush = false;
+    int width;
+    int height;
 
     bool used = false;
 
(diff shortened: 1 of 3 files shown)


jrfonseca:
I don't quite understand the problem or the proposed fix.

> If a user call glViewport with non default framebuffer bind first, and then bind the default framebuffer to draw, the current apitrace won't show anything on screen and complain glViewport hasn't been called yet. 

This should never happen.

When tracing, apitrace will emit a fake glViewport call the very first time a context is bound.  See [`_fake_glViewport()`](https://github.com/apitrace/apitrace/search?q=_fake_glViewport).

So if you do see such trace, then I suspect the problem actually happened when tracing, and the right fix is to ensure `_fake_glViewport()` is called the very first time the context is bound.

Also, `width` and `height` should be initialized to zero.

PR #762, opened 2021-09-21
retrace: Add option to run glReadPixels and glGetTexImage calls

gerddie:
These calls can be very expoensive, and for profiling it is of interest
to actually run them, in order to correctly identify performance
bottlenecks.


retrace/glretrace.py
```diff
@@ -171,9 +171,38 @@ def retraceFunctionBody(self, function):
             print(r'    if (currentContext && currentContext->features().pixel_buffer_object) {')
             print(r'        glGetIntegerv(GL_PIXEL_PACK_BUFFER_BINDING, &_pack_buffer);')
             print(r'    }')
-            print(r'    if (!_pack_buffer) {')
+            print(r'    if (!_pack_buffer && !retrace::runReadback) {')
             print(r'        return;')
             print(r'    }')
+            print(r'    if (retrace::runReadback) {')
```
jrfonseca, line 177: What if `_pack_buffer` is true here?  It would be wrong to do this.

retrace/glretrace.py
```diff
@@ -171,9 +171,38 @@ def retraceFunctionBody(self, function):
...
+            if function.name == "glGetTexImage":
+               print(r'     GLint max_tex_size;')
+               print(r'     glGetIntegerv(GL_MAX_TEXTURE_SIZE, &max_tex_size);')
+               print(r'     std::vector<char> buffer(max_tex_size * max_tex_size * max_tex_size);');
+               print(r'     glGetTexImage(call.arg(0).toUInt(), call.arg(1).toSInt(), call.arg(2).toUInt(), call.arg(3).toUInt(), &buffer[0]);');
```
jrfonseca, line 182: Instead of explicitly call `glGetXXX`, I'd simply allocate the output buffer, and let the automatically generated code to call `gl*` follow through.

That is, something like this:

```python
    print(r'    std::vector<char> buffer;')
    print(r'     if (!_pack_buffer) {')
    if function.name == "glGetTexImage":
        print(r'    buffer.resize( FIXME )')
        print(r'    pixels = &buffer[0]')
    elif function.name == "....":
        ...
    else:
        # Don't know how to handle this one, better return
        print(r'   return;')
    ....
    print(r'}')
    # The actual call will be generated later
```
But this needs to happen not here, but rather right before the call.

retrace/retrace_main.cpp
```diff
@@ -800,6 +802,7 @@ longOptions[] = {
     {"pdrawcalls", required_argument, 0, PDRAWCALLS_OPT},
     {"query-handling", required_argument, 0, QUERY_HANDLING_OPT},
     {"query-tolerance", required_argument, 0, QUERY_CHECK_TOLARANCE_OPT},
+    {"run-readbacks", no_argument, 0, RUN_READBACK_OPT},
```
jrfonseca, line 805: I'm actually OK not adding an option and do this unconditionally.

jrfonseca (commented):
This is a good idea.

Other than the implementation issues noticed, this LGTM.

retrace/glretrace.py
gerddie: Should be fixed with the refactoring.

PR #764, opened 2021-10-06
WIP Trace and replay ANGLE (EGL & GLES) on Windows

chrisdjali-wrld3d:
Hopefully this doesn't break anything else that already worked. It's also got a few other unrelated fixes for things that were causing problems during testing.

Although this doesn't work with the application I'd intended to use it with yet (due to a threading regression in ANGLE tracked as https://bugs.chromium.org/p/angleproject/issues/detail?id=6427), it should work with other applications compatible with vaguely up-to-date ANGLE versions. Hopefully that means it's useful to people and also reviewable. However, I'm seeing a problem with some other applications where the tracing functions are calling themselves instead of the originals which I've not go to the bottom of yet.

It uses the vcpkg port of ANGLE instead of the official Google version as Google offer no binaries and require DepotTools to build it, and then there's no `FindANGLE.cmake` or anything like that, so it would have been a hassle to do it any other way.


chrisdjali-wrld3d:
The recursion issue looks to be because of this feature of `LoadLibraryA`:

> If *lpFileName* does not include a path and there is more than one loaded module with the same base name and extension, the function returns a handle to the module that was loaded first.

The application I was originally working with loads ANGLE immediately as it's got the DLLs in its import table, whereas the others I tried load it manually later on. The injector injects the tracing DLL in between those events, so depending on the application, one or the other can be loaded first, and therefore returned by `LoadLibraryA`.

PR #765, opened 2021-11-09
retrace: Decode more attribute formats

gerddie:
Decode GL_INT_2_10_10_10_REV, GL_UNSIGNED_INT_2_10_10_10_REV, and
GL_UNSIGNED_INT_10F_11F_11F_REV attribute types.

### DIFF
--- a/retrace/glstate_shaders.cpp
+++ b/retrace/glstate_shaders.cpp
@@ -40,7 +40,7 @@
 #include "glstate.hpp"
 #include "glstate_internal.hpp"
 #include "halffloat.hpp"
-
+#include "floating_point_conversion.hpp"
 
 namespace glstate {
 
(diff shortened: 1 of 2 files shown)


retrace/glstate_shaders.cpp
```diff
@@ -304,6 +304,105 @@ struct VertexAttribDesc : AttribDesc
     }
 };
 
+/* code for converting fp11 and fp10 values to fp32 is taken from Mesa:
+ *
+ * The* Copyright (C) 2011 Marek Olšák <maraeo@gmail.com>
```
jrfonseca, line 309: It's the same license (MIT), so instead to duplicating the license, please just add `Copyright (C) 2011 Marek Olšák <maraeo@gmail.com>` to the top of this file, below Jose Fonseca.

Another alternative, is to put these helpers on their own header file.  Probably a good idea on its own right.

jrfonseca (commented):
Otherwise looks good. Thanks

retrace/glstate_shaders.cpp
gerddie: I did the latter and added a header file in helpers with these functions.

jrfonseca:
I believe this was arlready merged as part of https://github.com/apitrace/apitrace/pull/771 so closing.

Sorry for the delay.  (It's only now I'm on PTO I'm finding a bit of time to catch up.)

PR #766, opened 2021-11-09
egltrace: implement eglSetDamageRegion/eglSwapBuffersWithDamage on retrace

enunes:
These seem to be functional, although I'm not sure if this is the correct way to implement it in apitrace. Just let me know if something can be improved.

### DIFF
--- a/retrace/glws.hpp
+++ b/retrace/glws.hpp
@@ -148,6 +148,8 @@ public:
         visible = true;
     }
 
+    virtual void setDamageRegion(int *rects, int nrects) {}
+
     virtual void setName(const char *name) {}
 
     virtual void copySubBuffer(int x, int y, int width, int height);
--- a/retrace/glws.hpp
+++ b/retrace/glws.hpp
@@ -155,6 +155,8 @@ public:
     virtual void copySubBuffer(int x, int y, int width, int height);
 
     virtual void swapBuffers(void) = 0;
+
+    virtual void swapBuffersWithDamage(int *rects, int nrects) {}
 };
 
 
--- a/retrace/glws_egl_xlib.cpp
+++ b/retrace/glws_egl_xlib.cpp
@@ -228,6 +228,15 @@ public:
         eglSwapBuffers(eglDisplay, surface);
         processKeys(window);
     }
+
+    void swapBuffersWithDamage(int *rects, int nrects) override {
+        if (!checkExtension("EGL_KHR_swap_buffers_with_damage", eglExtensions))
+            return;
+
+        bindAPI(api);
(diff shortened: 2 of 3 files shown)


retrace/glws.hpp
```diff
@@ -148,11 +148,15 @@ class Drawable
...
 
+    virtual void setDamageRegion(int *rects, int nrects) {};
+
     virtual void setName(const char *name) {}
 
     virtual void copySubBuffer(int x, int y, int width, int height);
 
     virtual void swapBuffers(void) = 0;
+
+    virtual void swapBuffersWithDamage(int *rects, int nrects) {};
```
jrfonseca, line 159: This function should default to call `swapBuffers`.

`;` should be removed.

retrace/glws.hpp
```diff
@@ -148,11 +148,15 @@ class Drawable
         visible = true;
     }
 
+    virtual void setDamageRegion(int *rects, int nrects) {};
```
jrfonseca, line 151: `;` should be removed.

retrace/glws_egl_xlib.cpp
```diff
@@ -129,6 +129,13 @@ class EglDrawable : public Drawable
         eglWaitNative(EGL_CORE_NATIVE_ENGINE);
     }
 
+    void
+    setDamageRegion(int *rects, int nrects) override {
+        EGLint value;
+        eglQuerySurface(eglDisplay, surface, EGL_BUFFER_AGE_KHR, &value);
```
jrfonseca, line 135: You should check here that the respective EGL extension is supported, and do nothing if not.

retrace/glws_egl_xlib.cpp
```diff
@@ -218,6 +225,12 @@ class EglDrawable : public Drawable
         eglSwapBuffers(eglDisplay, surface);
         processKeys(window);
     }
+
+    void swapBuffersWithDamage(int *rects, int nrects) override {
+        bindAPI(api);
```
jrfonseca, line 230: Likewise, here one should check if the EGL extension is supported, and do nothing otherwise.

jrfonseca (commented):
You also need to look at the build failures. For WGL/CGL/etc, the best thing to do is to merely call standard swapBuffers.

Otherwise looks reasonable to me.

enunes:
All of those should be fixed now. I think WGL/CGL/etc are not affected since these are for EGL only. Apparently the CI errors on the windows build were related to the variable sized array.

enunes:
Ping, could you take another look at this one?

jrfonseca:
Looks good.  Thanks

PR #767, opened 2021-11-10
retrace: Add support for spoofing GetModuleFileNameW

Joshua-Ashton:
We use this in DXVK for app profiles so we need this to be spoofed like this too on retrace.


jrfonseca:
Looks good.  Thanks

PR #768, opened 2021-11-10
cmake: Always use XNAMath for MinGW

Joshua-Ashton:
I build on a case-sensitive file system, the DirectX math headers don't respect casing properly.

### DIFF
--- a/thirdparty/directxtex/CMakeLists.txt
+++ b/thirdparty/directxtex/CMakeLists.txt
@@ -7,7 +7,7 @@ if (DirectX_D3D11_4_INCLUDE_FOUND)
     )
 
     check_include_file_cxx (directxmath.h HAVE_DIRECTXMATH)
-    if (NOT HAVE_DIRECTXMATH)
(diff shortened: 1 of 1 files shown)


jrfonseca:
I don't care too much either way, but I wonder if this is self inflected somehow.   If the DirectX math headers you have don't respect casing, then how come they are in the include path?

jrfonseca:
To be clear, when building with MinGW, I recommend using DXSDK headers from https://github.com/apitrace/dxsdk

Joshua-Ashton:
Actually, looks like I don't have `directxpackedvector.h` at all from anywhere.
It's not in my MinGW headers.

The MinGW headers provide `directxmath.h` but not `directxpackedvector.h`. https://github.com/mirror/mingw-w64/tree/master/mingw-w64-headers/include

Looks like we may want to pull in from here: https://github.com/microsoft/DirectXMath

Joshua-Ashton:
Will resolve this in another MR.

PR #769, opened 2021-11-10
cmake: Fix linking against winpthreads

Joshua-Ashton:


jrfonseca:
I'm not entirely sure what side-effects `-static` might have, but it seems to cause no problems, so if it fixes your setup, fine by me.

PR #770, opened 2021-11-10
thirdparty: Use DirectXMath as a submodule

Joshua-Ashton:
Fixes building on modern MinGW which partially provides some older version.

Fixes code to compile with this under MinGW which has some weird define for no overloads on GCC for some reason.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -26,3 +26,6 @@
 	path = thirdparty/zlib
 	url = https://github.com/madler/zlib
 	shallow = true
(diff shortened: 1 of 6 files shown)


jrfonseca:
Sorry for the delay.  I hope this not only builds correctly on MinGW, but also works correctly, as I have a vague recollection of problems with certain versions of DirectXMath in the past.

I've just renamed `thirdparty/DirectXMath` to `thirdparty/directxmath` for consistency and pushed.  Thanks.

PR #772, opened 2021-11-24
gltrim: fix crash when invalid target in glActiveTexture

yuq:
Some application pass invalid target to glActiveTexture,
so glBindTexture crash with log:

Unknown bindpoint in call 3067:glBindTexture
gltrim: frametrim/ft_dependecyobject.cpp:202: frametrim::UsedObject::Pointer frametrim::DependecyObjectMap::bind(const trace::Call&, unsigned int): Assertion `0' failed.

### DIFF
--- a/frametrim/ft_dependecyobject.cpp
+++ b/frametrim/ft_dependecyobject.cpp
@@ -808,7 +808,36 @@ TextureObjectMap::TextureObjectMap():
 void
 TextureObjectMap::oglActiveTexture(const trace::Call& call)
 {
-    m_active_texture = call.arg(0).toUInt() - GL_TEXTURE0;
+    unsigned target = call.arg(0).toUInt();
+    switch (target) {
+    case GL_TEXTURE_CUBE_MAP_NEGATIVE_X:
+    case GL_TEXTURE_CUBE_MAP_NEGATIVE_Y:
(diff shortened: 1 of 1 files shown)


gerddie:
No, [glActivetexture](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glActiveTexture.xhtml) takes the *texture unit* not the *texture target*, so if the bind call gets a wrong target, then the reason for the assertion failure must be something
different.  I'd be happy to take a look if you have a trace that triggers this.

yuq:
Oh, right, I should check with texture unit number. But it has to be get from glGet(GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS) at runtime. Maybe assume a big enough number like 1024? In my case the invalid number is 4294933312 which is -GL_TEXTURE0.

Attach the trace which can trigger the probem: [catia.9486.trim.zip](https://github.com/apitrace/apitrace/files/7594662/catia.9486.trim.zip)

gerddie:
@yuq sorry for not responding earlier, I've added a different fix with https://github.com/apitrace/apitrace/pull/775. 
The trace you posted is strange in the sense that the calls to `glClientActiveTexture` always seem to use values that are completely off.

yuq:
OK, thanks, your fix works for me.

PR #775, opened 2022-01-21
Frametrim: sanititze texture unit in ActiveTexture calls and add more state function 

gerddie:
This fixes a crash that https://github.com/apitrace/apitrace/pull/772 tried to fix differently, and adds the functions that can be found in the trace attached there that were not yet handled.


jrfonseca:
Sorry for the delay, @gerddie   I've been busy personally (house renovation/move), and I hadn't notice there was a PR from you.

PR #790, opened 2022-05-13
gltrim: handle glFramebufferTexture3D and glMultiDrawElements

gerddie:


okias:
```
Tested-by: David Heidelberg <david.heidelberg@collabora.com>
```

PR #791, opened 2022-05-13
implement ARB_create_context_no_error for GLX and WGL

okias:
Fixes warnings as:
```
apitrace: warning: glXCreateContextAttribsARB: unknown key 0x31B3, interpreting value as int
```


jrfonseca:
LGTM.  Thanks.

Issue #786, opened 2022-04-28
Crash in _glGetIntegerv_override() with NULL config


ubitux:
Apparently in `_glGetIntegerv_override()`, `config` can be `NULL` causing the following alignment reading code to crash:

```c
        case GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT:
            if (config->ubo_offset_alignment > params[0])
                params[0] = config->ubo_offset_alignment;
            break;
        case GL_TEXTURE_BUFFER_OFFSET_ALIGNMENT:
            if (config->tbo_offset_alignment > params[0])
                params[0] = config->tbo_offset_alignment;
            break;
        case GL_SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT:
            if (config->ssbo_offset_alignment > params[0])
                params[0] = config->ssbo_offset_alignment;
            break;
```

Here is the backtrace:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f14405af71d in gltrace::_glGetIntegerv_override (pname=35380, params=0x7f142d657370)
    at /home/ux/src/apitrace/wrappers/glcaps.cpp:277
277	            if (config->ubo_offset_alignment > params[0])
[Current thread is 1 (Thread 0x7f142d658640 (LWP 102692))]
(gdb) bt
#0  0x00007f14405af71d in gltrace::_glGetIntegerv_override(unsigned int, int*)
    (pname=35380, params=0x7f142d657370) at /home/ux/src/apitrace/wrappers/glcaps.cpp:277
#1  0x00007f144042e4e2 in glGetIntegerv(GLenum, GLint*) (pname=35380, params=0x7f142d657370)
    at /home/ux/src/apitrace/build/wrappers/egltrace.cpp:18691
[...]
(gdb) p config
$1 = (const gltrace::configuration *) 0x0
```

jrfonseca:
This was introduced in 57e930496. @gerddie, access to `config` members need to be predicated by `config != nullptr`.  Can you fix this?

gerddie:
Ack, I'll take a look

PR #793, opened 2022-05-20
wrappers/glcaps: Check config availablility before dereferencing

gerddie:
Fixes: https://github.com/apitrace/apitrace/issues/786


jrfonseca:
Looks good. Thanks.

PR #795, opened 2022-05-23
scripts: add missing shebang for Python 3

okias:
Reported by lintian.

```
W: apitrace: executable-not-elf-or-script usr/lib/apitrace/scripts/highlight.py
```


jrfonseca:
Oh, I just noticed the rationale: lintian.

The script is not executable on the source tre:
```
$ ls -l scripts/highlight.py
-rw-rw-r-- 1 jfonseca jfonseca 6356 Feb 19  2021 scripts/highlight.py
```

If `usr/lib/apitrace/scripts/highlight.py` is executable, then the problem is the installation, which is wrongly adding the executable bit.

jrfonseca:
We need to fix `scripts/CMakeLists.txt` to install `highlight.py`not as as program...

okias:
Great catch. Thank you!

PR #807, opened 2022-06-23
gltrim: retain objects in the last frame that were created in earlier frames 

gerddie:
In order to be able to loop the last frame we must not destroy objects
that are used in the last frame but not created there. For simplicity
only check whether the object was created in the last frame, and not
emit the delete code if it was not.

Remark: 
This doesn't cover all cases where looping the last frame will fail, e.g. buffer mappings or context creation and destruction may also lead to crashes.


jrfonseca:
Looks reasonable to me.

PR #809, opened 2022-06-27
cli_info: implement ProcessName reporting

okias:
example output:
```
export TRACE=~/Downloads/unigi/Unigine_Heaven-4.0/bin/heaven_x64.trace
./apitrace info $TRACE
{
  "FileName": "/home/okias/Downloads/unigi/Unigine_Heaven-4.0/bin/heaven_x64.trace",
  "ProcessName": "/home/okias/Downloads/unigi/Unigine_Heaven-4.0/bin/heaven_x64",
  "ContainerVersion": 6,
  "ContainerType": "Snappy",
  "API": "OpenGL + GLX/WGL/CGL",
  "FramesCount": 11252,
  "ActualDataSize": 524912758,
  "ContainerSize": 345934639
}
```

When `glretrace` will be invoked as shown below:

```
export TRACE=~/Downloads/unigi/Unigine_Heaven-4.0/bin/heaven_x64.trace
(exec -a "$(./apitrace info ${TRACE} | jq '.["ProcessName"]')" ./glretrace ${TRACE})
```

Mesa3D library will see it under original application or game name, which
will retain drirc hacks needed for specific broken app/game to work
(thus for trace to correctly replay).

### DIFF
--- a/cli/cli_info.cpp
+++ b/cli/cli_info.cpp
@@ -81,6 +81,13 @@ getApiName(int api) {
   return trace::API_NAMES[api];
 }
 
+static const std::string
+getProcessName(std::string name) {
+  if (name == "")
+    return "null";
+  return "\"" + name + "\"";
+}
+
 struct FrameEntry {
     size_t firstCallId, lastCallId;
     size_t totalCalls;
@@ -91,6 +98,7 @@ static int
 command(int argc, char *argv[])
 {
     bool flagDumpFrames = false;
+    std::string processName;
     int opt;
     while ((opt = getopt_long(argc, argv, shortOptions, longOptions, NULL)) != -1) {
         switch (opt) {
@@ -118,6 +126,9 @@ command(int argc, char *argv[])
         if (!p.open(argv[i])) {
             return 1;
         }
+
+	auto &properties = p.getProperties();
+	std::string processName = properties.find("process.name")->second;
 
         trace::Call *call;
         size_t callsInFrame = 0;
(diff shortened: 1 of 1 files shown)


cli/cli_info.cpp
```diff
@@ -119,6 +120,9 @@ command(int argc, char *argv[])
             return 1;
         }
 
+	auto &properties = p.getProperties();
+	std::string processName = properties.find("process.name")->second;
```
jrfonseca, line 124: Old traces don't have process name (or any properties at all.)  You need to cope with that here.  Perhaps we can pass JSON's `null` in that case.

Otherwise LGTM.

okias: do you have some trace in old format laying around for me to test?

jrfonseca: https://github.com/apitrace/apitrace-tests/tree/master/traces has a few

cli/cli_info.cpp
```diff
@@ -81,6 +81,13 @@ getApiName(int api) {
   return trace::API_NAMES[api];
 }
 
+static const std::string
+getProcessName(std::string name) {
+  if (name == "")
```
jrfonseca, line 86: `if (name.empty)()`  , but per comment below, we can omit this function entilrely.

cli/cli_info.cpp
```diff
@@ -119,6 +127,9 @@ command(int argc, char *argv[])
             return 1;
         }
 
+	auto &properties = p.getProperties();
+	std::string processName = properties.find("process.name")->second;
```
jrfonseca, line 131: This still looks incorrect to me.

Here `properties` is a `std::map`, and `std::map::find` returns `end` iterator when nothing is found, and per https://en.cppreference.com/w/cpp/container/map/end  _"attempting to access it results in undefined behavior."_

Please use the standard `std::map::find` pattern for this.  That is, something along the lines of:
```c++
auto & it = properties.find("process.name");
if (it == properties.end()) {
    std::cout << "none";    
} else {
    std::cout << '\"' << it->second << '\"';
}
```

jrfonseca (changes requested):

PR #813, opened 2022-07-05
docs: update attachment procedure for trace

okias:
`apitrace repack -b` seems to be more effective than repacking with xz or lzma.


jrfonseca:
Makes sense.  Thanks.

jrfonseca (approved):

PR #815, opened 2022-07-25
Retrace: Add retrace watchdog flag

ddmail2009:
Add a flag to abort the trace if a single call takes too long.


jrfonseca:
Looks useful.  Thank you!

PR #816, opened 2022-08-05
Fix bug in _EGLImageKHR_get_image_info

per-mathisen-arm:
It is currently creating a texture then attempting to delete it as a buffer.


jrfonseca:
Great find.  Thanks!

Issue #820, opened 2022-08-31
gltrim doesn't generate anytrace and doesn't show any error..


oscarbg:
Hi,
wanted to make a small trace to provide a bug report for Zink.. after being big size I found about gltrim -k to save a frame only which should be enough..
https://www.collabora.com/news-and-blog/blog/2021/02/01/trimming-apitrace-workload-captures-for-better-mesa-testing/
sadly seems command doesn't generate any file and doesn't show any error..
file is 200MB available the bug report: (from https://gitlab.freedesktop.org/mesa/mesa/-/issues/7087#note_1528557)
doing a simple:
`apitrace gltrim -f 10  -o b.trace frames4k.trace`
exhibits the issue..
at least I would expect a trace to be generated even I replaying it wouldn't generate a correct frame due to some missing state not being saved correctly..
adding @gerddie as it seems to be the main author of the feature..
EDIT: using latest Win github release (apitrace-11.1-win64) as appveyor builds link don't work #818

gerddie:
Trimming the trace leads to a sigsegv on Linux, the error doesn't show up when running by using the `apitrace`, but it can be seen when running gltrim directly. My first guess is that the crash is triggered by the wgl calls not being handled.

gerddie:
There were actually a few more problems, namely unhandled direct state access calls, and soemhoe the trace seems to give out query IDs incorrectly, which made me remove an assert and just print out a warning. The trace becomes gl-trimmable with this: 

https://github.com/apitrace/apitrace/pull/821

gerddie:
BTW: (a)To just cut of the tail of the trace you could also have used `apitrace trim --frames 0..10 -o trim.trace frames4.trace`. 
(b) I see a lot of errors when replaying the original trace (or the one that is trimmed like above). The gltrim trace doesn't retain the calls that result in these error messages.

PR #821, opened 2022-09-02
Submitted handle wgl and more gl45

gerddie:
This makes it possible to trim https://github.com/apitrace/apitrace/issues/820


jrfonseca:
Looks AFAICT

PR #836, opened 2022-11-12
d3d8trace: Remap shader handles

AlpyneDreams:
D3D8 uses DWORD handles instead of shader objects. The specific values of these handles is not defined, so it is best to remap them like we do with GL handles.

This fixes issues where a D3D8 trace will cause errors when run with [d8vk](https://github.com/AlpyneDreams/d8vk) or vice-versa.

### DIFF
--- a/specs/d3d8.py
+++ b/specs/d3d8.py
@@ -77,7 +77,11 @@ HRESULT = MAKE_HRESULT(ok = "D3D_OK", errors = [
     "D3DERR_DRIVERINVALIDCALL",
 ])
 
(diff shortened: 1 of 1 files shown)


jrfonseca:
LGTM.  Thanks!

jrfonseca:
Change looks perfect.  Thanks for taking care of this!

dxvk8 looks a cool project.  Another D3D8 program we used a lot back in the day was 3DMark2001 from https://benchmarks.ul.com/legacy-benchmarks

PR #844, opened 2022-12-06
d3d8trace: Remap state block handles

AlpyneDreams:
Similar to #836, D3D8 uses DWORD handles instead of state block objects. We have to track them like GL handles.

Similarly, this fixes crashes when replaying traces on alternate D3D8 implementations such as [d8vk](https://github.com/AlpyneDreams/d8vk).


jrfonseca:
Thanks

Issue #848, opened 2022-12-16
GL_ATI_meminfo state: 4 integers instead of 1


stgatilov:
While trying to glretrace my game on AMD RX 550 with VS2022 x64 Debug, I get stack corruption around `texture_free_memory_ati` variable:
```
        // GL_TEXTURE_FREE_MEMORY_ATI
        {
            flushErrors();
    GLint texture_free_memory_ati = 0;
    glGetIntegerv(GL_TEXTURE_FREE_MEMORY_ATI, &texture_free_memory_ati);
            if (glGetError() != GL_NO_ERROR) {
                flushErrors();
            } else {
                writer.beginMember("GL_TEXTURE_FREE_MEMORY_ATI");
    writer.writeInt(texture_free_memory_ati);
                writer.endMember();
            }
        }
```

Looking at the [specs for GL_ATI_meminfo extension](https://registry.khronos.org/OpenGL/extensions/ATI/ATI_meminfo.txt), glGetIntegerv returns 4 integers instead of 1 for all three states.

stgatilov:
Here is the fix: https://github.com/apitrace/apitrace/pull/849

jrfonseca:
Good catch.  Thanks

PR #852, opened 2022-12-18
#848. Check for buffer overflow in case of scalar glGet[Integer]v.

stgatilov:
This is useful in case we (or implementation) misunderstood specs and read scalar where it is an array.


jrfonseca:
Thanks

PR #856, opened 2023-01-18
GL_ARB_cl_event: workaround via glFenceSync during replay

stgatilov:
When replaying glCreateSyncFromCLeventARB, make a typical call to glFenceSync instead.
That would create a sync object which can be waited for and deleted, so from OpenGL view it behaves exactly as desired.
(Of course it does not perform any synchronization with OpenCL context since we don't record/replay OpenCL)


stgatilov:
Not sure in what I should pass to trace::FunctionSig and trace::Call constructors.

PR #863, opened 2023-02-10
gltrim: Add option to replace swaps in setup with glFinish

gerddie:
In order to not accumulate too much work in the setup frame add the option to replace the buffers swaps that would be happening there with glFinish to submit the work to the graphics card. This is useful for the Mesa-CI.


anholt:
I had been thinking glFlush before, but glFinish is even better, since it'll be closer to the throttled behavior of the swaps we're dropping.  This is great.

jrfonseca:
Makes sense. LGTM.

PR #864, opened 2023-03-08
Add missing calls to handle the GLES version of GravityMark 

gerddie:
* Add support for a few more GLES calls 
* Take glBufferStorageEXT into account when handling shadow buffers 
* Add calls to gltrim that are used in a trace obtained by running the GLES version of the benchmark 

Remark: It seems something is missing with the buffer storage handing on GLES: When I trace the benchmark with EXT_buffer_storage enabled, then no asteroids are drawn (Intel IRIS GPU), when I disable EXT_buffer_storage, the benchmark runs correctly and trace is fine, when just running the benchmark (no tracing) all is fine too.


jrfonseca:
Looks perfect.  Thanks

PR #866, opened 2023-03-28
cmake: enable LTO

okias:
```
444M	build
360M	build-lto
```

TBD:
 - test replay performance (seems to be good)
 - fix GCC / Clang warnings

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,9 @@ project (apitrace)
 # command so that we can default to auto-detecting optional depencies, while
 # still providing a mechanism to force/disable these optional dependencies, as
 # prescribed in http://www.gentoo.org/proj/en/qa/automagic.xml
+if (NOT MINGW)
+    set (CMAKE_INTERPROCEDURAL_OPTIMIZATION True)
+endif ()
 set (ENABLE_GUI "AUTO" CACHE STRING "Enable Qt GUI.")
 option (ENABLE_QT6 "Use Qt6 support." OFF)
 
--- a/lib/trace/trace_parser.cpp
+++ b/lib/trace/trace_parser.cpp
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <climits>
 #include <memory>
 
 #include "trace_file.hpp"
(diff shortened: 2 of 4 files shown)


jrfonseca:
It looks like backtrace btest fails with LTO.  

You might be able to avoid the issue by disabling the `INTERPROCEDURAL_OPTIMIZATION` property on the `btest` target, and possibly backtrace library itself.

jrfonseca:
What I suggested is just incrementally different, not radically different, as `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is used to initalize `INTERPROCEDURAL_OPTIMIZATION` properties.

To be clear, what I'm suggesting is `set (CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)`  globally (as you already do), then `set_target_properties(foo PROPERTIES INTERPROCEDURAL_OPTIMIZATION FALSE)` where and only if needed.

okias:
Had to switch the LTO off for whole libbacktrace, since disabling per target (btest/stest) didn't helped. Anyway the results is good (in terms of size).

jrfonseca:
https://github.com/apitrace/apitrace/pull/866/commits/db3c2c5ec5bf68b8362776154cfd940dc912434d looks good.

https://github.com/apitrace/apitrace/pull/866/commits/7b8d60c339f312e992eab03f56a506a97e1a22d3 surprised me.  IIUC, from https://developers.redhat.com/blog/2017/02/22/memory-error-detection-using-gcc description `-Walloc-size-larger-than=` warning relies on static code analysis.  I'd prefer an `assert`, but likely the warning would persist on non-debug builds, so I suppose clamping is indeed the simplest thing to do.  I think we should use `PTRDIFF_MAX` instead of `INT_MAX`, to avoid unnecessarily limits on 64-bits builds.  Otherwise that too looks good.

okias:
Thanks, updated to the `PTRDIFF_MAX`, good catch :)

okias:
hmm, the `ubuntu32` job is broken :(

jrfonseca:
It's not your fault.  GitHub broken the ubuntu base image for 32bit packages somehow.  If you rebase, it should pick up master change to use docker for this job, and it should start working,

okias:
Thanks, rebased.

PR #869, opened 2023-04-28
wrappers: add noexcept specification to DWriteCreateFactory

lfrb:
The `winnt.h` header seems to have changed recently and there is a `WIN_NOEXCEPT` exception specification added to `DWriteCreateFactory`. It is either defined as `noexcept` or `throw()` depending on the compiler version. If `WIN_NOEXCEPT` is not defined, it means we have an older version of the header, so we define it as empty.


jrfonseca:
Looks good.  Thanks

PR #870, opened 2023-04-28
frametrimmer: split OpenGL implementation into a new class

lfrb:
To allow implementations for different graphics API, the OpenGL implementation has been moved to a new class (`frametrimmer::OpenGLImpl`). The `FrameTrimmer` class now only contains the most generic methods and callback caching.


gerddie (approved):
Apart from the little issue the series looks fine to me. R-b

PR #876, opened 2023-07-02
specs: Add support for EGL_EXT_image_dma_buf_import_modifiers

linkmauve:
This extension allows programs to use efficient memory layouts.

Ideally both formats and modifiers would get displayed as enums, or at least as hex, but I’m not sure it makes sense to copy the full `drm_fourcc.h` in there.

I have tested this commit against both Weston and [Catacomb](https://github.com/catacombing/catacomb), two Wayland compositors.


linkmauve:
Superseded by #877.

PR #877, opened 2023-07-02
Implement a bunch of additional EGL extensions

linkmauve:
All of these are used or at least linked to by Wayland compositors, this makes it much easier to debug what is going on while debugging those, as otherwise the function calls just don’t get displayed in `apitrace dump`.

I have tested those against both Weston and Catacomb, on iris, lima and nouveau.

This series is more easily read per commit than as a whole.

I’m not quite sure how to best handle the third commit, which would look much better if I would basically import `drm_fourcc.h`, but then keeping it up to date might be an issue.

Supersedes #876.


jrfonseca:
Looks good AFAICT.  Thanks.

jrfonseca:
> I’m not quite sure how to best handle the third commit, which would look much better if I would basically import `drm_fourcc.h`, but then keeping it up to date might be an issue.

I have no issues with the commit as is, as it doesn't break anything. 

Including `drm_fourcc.h` might be a hassle, as it might not be always present, especially outside Linux.  The best is to duplicate the enum.

Another solution is to wrap the FourCC codes with `trace::Repr`, which allows to give a human legible representation to any value.  You can see an example of using this to wrap Direct3D shader byte code with a human legible disassembly on https://github.com/apitrace/apitrace/blob/master/wrappers/d3dcommonshader.cpp

PR #880, opened 2023-08-02
Work around Qt6 crash involving LTO

keithel:
If link time optimization is enabled in a Qt 6 build of qapitrace, signals are not properly connected, and a crash occurs.

Somehow an interaction between the `image` library and Qt, along with the `INTERPROCEDURAL_OPTIMIZATION` CMake setting causes this behavior.

Simply disabling link time optimization by setting the `INTERPROCEDURAL_OPTIMIZATION` qapitrace target property to false fixes the issue.

### DIFF
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -97,6 +97,10 @@ add_dependencies(apitrace version)
 
 add_executable (qapitrace ${qapitrace_SRCS} ${qapitrace_UIS_H})
 
+# Turn off Linktime Optimization because it causes issues with Qt signal slot connections.
+# See https://bugreports.qt.io/browse/QTBUG-115731
+set_property(TARGET qapitrace PROPERTY INTERPROCEDURAL_OPTIMIZATION FALSE)
+
 target_link_libraries (qapitrace
     Qt::Widgets
     Qt::Network


okias:
I would wait a bit with merging something like this, since other possible solutions are proposed in https://bugreports.qt.io/browse/QTBUG-115731 . 

In worst case it would be nice if it would be guarded by Qt 6 checks.

keithel:
Actually -- I get this even with Qt5 builds with the compiler, linker and glibc I use.  But yeah, I can see waiting until we know more about the underlying issue. It's likely this will just be closed on the Qt side as a compiler issue.

keithel:
@okias could you verify that my proposed fix works for you (avoids the crash, and allows signals to be connected properly, and the UI to be functional) ?

okias:
I would prefer the workaround to be applied only where necessary, such as detecting or defining broken configurations. If it's not easily possible, then I suppose I would be okay with it as it is.

keithel:
It sounds like this may be fixed in Qt by some changes in the cmake rules, as it appears that adding -fPIC to the linker line fixes the problem - according to Alexandru Croitor. I think we just let this PR sit until QTBUG-115731 is fixed or otherwise closed. I'll make narrower changes when it is more clear as to what to do.

PR #882, opened 2023-08-05
Improve the layout of the retracer dialog

keithel:
The retracer dialog had a lot of weird things going on. This should improve it so it looks right on all platforms.

### DIFF
--- a/gui/ui/retracerdialog.ui
+++ b/gui/ui/retracerdialog.ui
@@ -28,9 +28,6 @@
    </property>
    <item>
     <layout class="QVBoxLayout" name="checkboxesLayout">
-     <property name="leftMargin">
-      <number>27</number>
-     </property>
      <item>
       <widget class="QCheckBox" name="doubleBufferingCB">
        <property name="text">
--- a/gui/ui/retracerdialog.ui
+++ b/gui/ui/retracerdialog.ui
@@ -9,8 +9,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>314</width>
-    <height>300</height>
+    <width>238</width>
+    <height>282</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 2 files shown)


okias:
Nice! Can you do screenshot before and after?

keithel:
Before, without any resizing:
image

Before, resizing the window manually, so you can see the whole content, and see how it expands unintelligently:
image


image

keithel:
The extra width in the dialog is such that the combobox is wide enough to fit the contents of all entries.
There is some extra width in the combobox beyond that, which I cannot figure out how to get rid of.

gui/ui/retracerdialog.ui
```diff
@@ -23,104 +23,28 @@
...
-        <size>
-         <width>20</width>
-         <height>20</height>
-        </size>
-       </property>
-      </spacer>
-     </item>
+    <layout class="QVBoxLayout" name="checkboxesLayout">
+     <property name="leftMargin">
+      <number>27</number>
```
okias, line 32: preferably number which could be divided by two

keithel: I set this to 27 so that pixel-for-pixel this matched what previously achieved with a horizontal spacer.
I can certainly adjust this though.

keithel: Honestly, not having this custom at all would look better, imho. So - making the left margin 0 (default).
I'll get you a screenshot of what that looks like as soon as I figure out this weird text elision bug.
the & in `Run & Check` is being elided to `Run ... Check` for some reason, when it shouldn't be.
The next commit you see is not intended to be committed, but is there to illustrate the issue in a Qt forum post.

I'll let you know when I think this is ready.

keithel:
Ok -seems that the "Run ...Trace" issue I am seeing is a bug that was introduced in Qt 5.15.3, never fixed in Qt 5.15, but fixed in Qt 6 sometime before Qt 6.5.1. I'll just assume this is not an issue.

gui/ui/retracerdialog.ui
keithel: Here's what it looks like without a left margin:

image

okias: It looks more minimalistic; I like it.

okias (approved):

keithel:
@okias does this need review from jrfonseca and/or gerddie to be merged?

okias:
On my configuration I don't see huge differences, but it looks a bit better then previous window which just has bigger and uglier placement.

keithel:
Yes - no radical changes - It was just ugly and some of the content was obscured when expanded or shrunk. Small change, intro to me making changes in this repo.

PR #893, opened 2023-09-06
retrace: call waffle_destroy_window on drawable destructor

italonicola:
Without this patch, replaying an eglDestroySurface() call won't generate a eglDestroySurface() call using the waffle backend, leaving any created windows alive indefinitely.

### DIFF
--- a/retrace/glws_waffle.cpp
+++ b/retrace/glws_waffle.cpp
@@ -69,6 +69,10 @@ public:
         window (win)
     {}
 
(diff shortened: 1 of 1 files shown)


okias:
LGTM, any sample of trace where it helps? (from https://gitlab.freedesktop.org/gfx-ci/tracie/traces-db/ ?)

okias:
Thank you!

PR #895, opened 2023-09-22
Allow building using system versions instead of thirdparty submodules

### LINKED ISSUE #891: Build system complains about uncloned submodules for bundled dependencies even when using dependencies from system
First system dependencies are searched (which is good):

https://github.com/apitrace/apitrace/blob/00706d1a395ab6570ab88136fffb3028a3ded3bd/CMakeLists.txt#L535-L562

Then submodules must exist regardless (which is bad):

https://github.com/apitrace/apitrace/blob/00706d1a395ab6570ab88136fffb3028a3ded3bd/thirdparty/CMakeLists.txt#L19-L36

But then most of those folders can probably be skipped because they are only used conditionally (which is good):

https://github.com/apitrace/apitrace/blob/00706d1a395ab6570ab88136fffb3028a3ded3bd/thirdparty/CMakeLists.txt#L38-L66

spillner:
Resolves #891.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -638,6 +638,7 @@ add_custom_target (version
         ${CMAKE_COMMAND} -D SRC=${CMAKE_SOURCE_DIR}/version.h.in
                          -D DST=${CMAKE_BINARY_DIR}/version.h
                          -D GIT_EXECUTABLE=${GIT_EXECUTABLE}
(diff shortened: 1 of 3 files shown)


okias:
The idea looks good to me, could you do "full package"? https://udd.debian.org/patches.cgi?src=apitrace&version=11.1%2Brepack-1.2

For example if I drop snappy from submodules, it won't allow me replace it with system package as a default, so if you could cleanly incorporate these patches, that would be awesome.

PR #899, opened 2023-09-29
fix #892

### LINKED ISSUE #892: Bug recording instanced draw calls with user arrays
Both traces provided in https://gitlab.freedesktop.org/mesa/mesa/-/issues/9119 crash during vao data upload for a draw. It seems like somehow either the vertex data is being truncated (it's awfully suspicious that they all have the same data size?) or the attribute divisor is being lost.

zmike:
Not sure if I should squash this or what, but it seems to resolve all issues with #892


jrfonseca:
Rebased and pushed.  Thanks

PR #902, opened 2023-10-16
gltrim: Fix trimmimg multi-threaded traces and fix a few calls 

gerddie:
* Fix CompressedTexture calls 
* Add support for EXT_framebuffer_object 
* Handle multiple sub-contexts in multiple threads 
* Handle a few more calls


jrfonseca:
LGTM.

PR #904, opened 2023-12-20
Fake command line arguments during replay, so they match traced program's arguments

DziubanMaciej:
Some middleware libraries and/or drivers may detect applications based on executable name and command line. Apitrace currently saves executable name in the trace and then injects its own implementation of some syscalls during replay to return this name.

This PR adds this functionality to command line to further enhance the illusion of replayed app being the real app. Currently only fully implemented on Windows.

### DIFF
--- a/retrace/process_name.cpp
+++ b/retrace/process_name.cpp
@@ -123,6 +123,10 @@ setProcessName(const char *processName)
     *p__progname = progname;
 }
 
+void
+setProcessCommandLine(const char* processCommandLine)
+{
+}
 
 #elif defined(_WIN32) && (defined(__i386__) || defined(__x86_64__) || defined(_M_IX86) || defined(_M_AMD64))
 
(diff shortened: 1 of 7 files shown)


retrace/process_name.cpp
```diff
@@ -250,6 +255,57 @@ setProcessName(const char *processName)
...
+{
+    return g_processCommandLine;
+}
+
+static LPWSTR WINAPI
+MyGetCommandLineW()
+{
+    static wchar_t processCommandLineW[4097];
+    static bool bConverted = false;
+    if (!bConverted)
```
jrfonseca, line 269: Please use style
```c++
    if (condition) {
       ....
    }
````
here and elsewhere.

jrfonseca (changes requested):
Otherwise looks good.

Please rebase on top of 97fd014cd48a7e0b6923373dbb08397309d6f669 to allow the the build to run without running into the Arm build failure.

jrfonseca:
Thank you

PR #905, opened 2023-12-31
minor fixes and additions

Slaw6820:
During my work with apitrace I have encountered a small number of issues that were addressed with the following changes to the source code.
Please review.

### DIFF
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -142,9 +142,9 @@ if (WIN32 OR APPLE OR X11_FOUND)
                 #"-framework OpenGL" # CGL*
             )
 
-            # Bundle Info.plist.  Set sdk version to 10.13.0 to avoid blank screen on Mojave
+            # Bundle Info.plist
             set_target_properties (glretrace PROPERTIES
-                LINK_FLAGS "-sectcreate __TEXT __info_plist ${CMAKE_CURRENT_SOURCE_DIR}/Info.plist -Wl,-sdk_version,10.13.0"
+                LINK_FLAGS "-sectcreate __TEXT __info_plist ${CMAKE_CURRENT_SOURCE_DIR}/Info.plist"
                 LINK_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/Info.plist"
             )
         else ()


okias:
Hello @Slaw6820 

thank you for the contribution, could you split the one commit into the separate changes?

Slaw6820:
Sure. Does it mean that all proposed fixes are accepted?
Should I create multiple changes under one PR or separate PRs?

okias:
> Sure. Does it mean that all proposed fixes are accepted? Should I create multiple changes under one PR or separate PRs?

LTGM, but I'm not an biggest authority here :) just commits within this MR :)

Slaw6820:
@jrfonseca

helpers/glfeatures.hpp
```diff
@@ -99,19 +99,19 @@ struct Profile {
...
 
     // Comparison operator, mainly for use in std::map
     inline bool
     operator < (const Profile & other) const {
-        return major < other.major ||
-               minor < other.minor ||
-               api < other.api ||
-               core < other.core ||
-               forwardCompatible < other.forwardCompatible;
+        return (major  < other.major) ||
```
jrfonseca, line 110: Good catch.

There's a more compact way to achieve the same:
```c++
        if (major != other.major) {
            return major < other.major;
        }
        if (minor != other.minor) {
            return minor < other.minor;
        }
        ...
````

File: frametrim/ft_opengl.cpp [hunk 2]
```diff
@@ -221,7 +221,16 @@ OpenGLImpl::skipDeleteObj(const trace::Call& call)
         map = &m_textures;
 
     if (!strcmp(call.name(), "glDeleteFramebuffers"))
-        map = &m_current_context->m_fbo;
+    {
```
jrfonseca, line 224: Code formatting should be
```c++
    if (condition) {
        ....
    }
````

File: retrace/retrace_swizzle.cpp [hunk 3]
```diff
@@ -32,6 +32,19 @@
 #include "retrace_swizzle.hpp"
 
 
+#ifdef assert
+#undef assert
+#endif
+
+
+#define assert(expression) \
```
jrfonseca, line 40: Why is this necessary?

jrfonseca (commented):
Sorry for the silence -- I was on vacation last week.

Thanks for the changes.  Comments inline.

Otherwise looks good.   I agree it would be good to break the changes according to main _modules_ (wrappers, retrace, etc.  The main reason is that one can git bisect and revert chunks

Issue #493, opened 2016-10-14
Unable to trace Dying Light


ghost:
# Introduction

apitrace 7.1 and apitrace-git fail with [Dying Light](http://store.steampowered.com/app/239140/)
# apitrace 7.1
- qapitrace is showing just the first two frames and is printing:

```
Loading: "DyingLightGame.bin.1.trace"
warning: unexpected end of file while reading trace
warning: unexpected end of file while reading trace
```
- "glretrace -d" aborts and prints:

```
1082: message: shader compiler issue 1: Shader Stats: SGPRS: 24 VGPRS: 16 Code Size: 64 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0                                                                                           
1082: message: shader compiler issue 1: Shader Stats: SGPRS: 24 VGPRS: 16 Code Size: 80 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0                                                                                           
1100: message: shader compiler issue 1: Shader Stats: SGPRS: 24 VGPRS: 16 Code Size: 224 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0                                                                                          
1100: message: shader compiler issue 1: Shader Stats: SGPRS: 24 VGPRS: 16 Code Size: 132 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0                                                                                          
warning: unexpected end of file while reading trace
Rendered 1 frames in 0.0959289 secs, average of 10.4244 fps
*** Error in `glretrace': free(): invalid size: 0x00007f4b75991010 ***
...
apitrace: warning: caught signal 6
3922: error: caught an unhandled exception
glretrace+0x20f9b4
/lib64/libpthread.so.0+0x1068f: /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/nptl/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
/lib64/libc.so.6: __GI_raise+0x39: ../sysdeps/unix/sysv/linux/raise.c:55
/lib64/libc.so.6: __GI_abort+0x16f: /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/stdlib/abort.c:89
[...]

PR #906, opened 2024-02-03
Fix tracing and replaying of Dying Light

speles:
Fixes https://github.com/apitrace/apitrace/issues/493

The fixes are in the recent code, so the issue is probably not the exact same as the original reporter had. But tracing and replaying the Dying Light was broken in 11.1 and current master too.

With these patches, I can confirm that both tracing and replaying actually works.

### DIFF
--- a/lib/os/os_posix.cpp
+++ b/lib/os/os_posix.cpp
@@ -146,11 +146,13 @@ getProcessCommandLine(void)
             size_t start = strlen(buf) + 1;
             size_t cmdlineLen = len - start;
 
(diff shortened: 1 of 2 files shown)


jrfonseca:
Hi @speles .  Thanks for your contribution.  This PR got lost in my inbox somehow -- I only noticed it now.  In the meanwhile, both issues were been fixed independently.  I sincerely apologize, to you whom fixed these issues first (and accurately!), and the others who ended up replicating your work unnecessarily...  I'll try to do better next time.

PR #914, opened 2024-04-25
Add option to append timestamp to generated trace filename

lfrb:
This is useful when traces are generated in an automated fashion or handled by some manager auto-restarting the process. This gives more information about the different trace files generated.

`TRACE_TIMESTAMP` environment variable can be used when using the wrappers directly without using the CLI to start tracing.

### DIFF
--- a/lib/trace/trace_option.cpp
+++ b/lib/trace/trace_option.cpp
@@ -42,7 +42,7 @@ boolOption(const char *option, bool default_) {
         strcmp(option, "false") == 0) {
         return false;
     }
-    if (strcmp(option, "0") == 0 ||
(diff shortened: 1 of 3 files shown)


lib/trace/trace_option.cpp
```diff
@@ -42,7 +42,7 @@ boolOption(const char *option, bool default_) {
         strcmp(option, "false") == 0) {
         return false;
     }
-    if (strcmp(option, "0") == 0 ||
```
jrfonseca, line 45: Good catch!

jrfonseca (approved):
Looks good.

PR #915, opened 2024-04-25
Display format modifiers as hexadecimal values

lfrb:
It doesn't make sense to show modifiers as a decimal value and they are easier to parse with the hexadecimal representation. 

This add a new API type (`Formatted`)  that is serialized as a `Repr` value in the trace file. The values are gonna show up properly as long as the newest version of apitrace (with these patches) is used to generate the trace as the human value is embedded in the file.

### DIFF
--- a/specs/eglapi.py
+++ b/specs/eglapi.py
@@ -61,6 +61,8 @@ EGLDeviceEXT = Opaque("EGLDeviceEXT")
 
 # EGL_EXT_image_dma_buf_import_modifiers
 EGLuint64KHR = Alias("EGLuint64KHR", UInt64)
+EGLModifier = Bitmask(EGLuint64KHR, [])
+Hex = Bitmask(Int, [])
 
 # EGL_KHR_image_base
 EGLImageKHR = Alias("EGLImageKHR", EGLImage)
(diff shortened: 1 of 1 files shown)


jrfonseca:
I think a new `Formatted` is overkill.  `Bitmask`s are already dumped as hexadecimals.

Please try an empty `Bitmask` instead, like:
```diff
diff --git a/specs/eglapi.py b/specs/eglapi.py
index 31e51cd9..0888f783 100644
--- a/specs/eglapi.py
+++ b/specs/eglapi.py
@@ -61,6 +61,8 @@ EGLDeviceEXT = Opaque("EGLDeviceEXT")
 
 # EGL_EXT_image_dma_buf_import_modifiers
 EGLuint64KHR = Alias("EGLuint64KHR", UInt64)
+EGLModifier = Bitmask(EGLuint64KHR, [])
+Hex = Bitmask(Int, [])
 
 # EGL_KHR_image_base
 EGLImageKHR = Alias("EGLImageKHR", EGLImage)
```

The only caveat is that `Bitmask` assumes the value is unsigned.  Hopefully it doesn't matter here.

jrfonseca:
Thank you.

PR #920, opened 2024-06-06
Some fixes related to the MRs I opened earlier

lfrb:
- Fix "-t" short option
- Don't mutually exclude "--timestamp" and "--output" options
- Dump the enum value name for compression rates instead of the value


chealy:
+1 on these fixes. Quite helpful.

jrfonseca:
Thanks

PR #923, opened 2024-06-28
Fix replay issue of glGetTextureHandle and glProgramUniformHandle:

baozholu:
Fix replay issue of glGetTextureHandle and glProgramUniformHandle:
    The return handle of glGetTextureHandleARB will differ between different vender/ASIC GPUs, current apitrace uses a _textureHandle_map and _imageHandle_map to store the original handle (Obtained during trace) and the new handle (Obtained during replay), but apitrace still uses the original handle when replaying the glUniformHandleui64ARB call, it should find the new handle through the _textureHandle_map and _imageHandle_map, then use it in glUniformHandleui64ARB. 
    For other APIs in the GL_ARB_bindless_texture extension, they used the _textureHandle_map/_imageHandle_map to store/get the real handle during replay, so we should find the new handle through the _textureHandle_map/_imageHandle_map and use it in glUniformHandle and glProgramUniformHandle, just like glMakeTextureHandle and glMakeImageHandle does All of the following APIs have this issue: glUniformHandle*, glProgramUniformHandle*

### DIFF
--- a/retrace/glretrace.py
+++ b/retrace/glretrace.py
@@ -347,6 +347,46 @@ class GlRetracer(Retracer):
             print(r'        retrace::warning(call) << " renderbuffer targets unsupported (https://git.io/JOMRC)\n";')
             print(r'    }')
 
+        if function.name in('glUniformHandleui64ARB', 'glProgramUniformHandleui64ARB', 'glUniformHandleui64NV', 'glProgramUniformHandleui64NV'):
+            print(r'    if (_textureHandle_map.find(value) != _textureHandle_map.end())')
+            print(r'    {')
+            print(r'        value = _textureHandle_map[value];')
+            print(r'    }')
+            print(r'    else if(_imageHandle_map.find(value) != _imageHandle_map.end())')
+            print(r'    {')
+            print(r'        value = _imageHandle_map[value];')
+            print(r'    }')
+
+        if function.name in('glUniformHandleui64vARB', 'glUniformHandleui64vNV'):
+            print(r'    const trace::Array* handleInfo = (call.arg(2)).toArray();')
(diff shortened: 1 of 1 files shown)


lubaozho:
@jrfonseca could you help to review it ?

jrfonseca:
Sorry, I need to find a bit of spare time to actually test this PR.  Because I still don't fully understand the problem and fear this change might interfere with other things.

lubaozho:
**Simplified this PR description:** 
glUniformHandle* and glProgramUniformHandle* should use the **_textureHandle_map/_imageHandle_map** to store/get the real handle during replay, just like glMakeTextureHandle and glMakeImageHandle does. All of these API belong to the GL_ARB_bindless_texture extension.

@jrfonseca can this PR be merged?

jrfonseca:
@baozholu / @baozholu-AMD ,

I apologize it took this long.  I had this on my todo list this time, and finally put some time.

I also finally understand the problem: `glUniformHandle`/`glProgramUniformHandle` take `GLuint64` as last args, but really should be taking the _union_ of a `GLtextureHandle`, `GLimageHandle`.

There might be a _cleaner_ way to deal with this, by using `glGetActiveUniform`/`glGetUniformLocation` to determine the type of uniform at the specified location, and use the appropriate handle table.

Another solution is to use a _single_ table instead of `GLtextureHandle` and `GLimageHandle`.  If these texture/image handles are GPU virtual addresses, then that's the most sensible thing to do.

That said, no way this change can introduce regressions, as no swizzling was happening at all, which was wrong.   So I'll merge as is.  `GLtextureHandle` and `GLimageHandle`

I'll merge `GLtextureHandle` and `GLimageHandle` in a subsequent change.

Thanks for you patience.

Issue #689, opened 2020-11-07
OpenGL ES profiling support


ammen99:
Hi all,

I have an OpenGL ES application that I want to profile. I have read that I need `--pcpu --pgpu --ppd` however:

- `--pgpu` requires `GL_ARB_timer_query` which is not available on OpenGL ES at all.
- `--ppd` requires `GL_ARB_occlusion_query` which is not available on OpenGL ES either.

Am I missing something or this means that the only thing possible to measure for OpenGL ES is the CPU time? (with `--pcpu`)

jrfonseca:
Sorry for the silence.  I really haven't been able to keep up with the inflow of issues and PRs on Apitrace.  It's probably too late to be of assistance, but better late than never...

It looks you're right.  OpenGL ES is missing those extensions, so it doesn't work.

ES doesn't seem to be anything resembling `GL_ARB_occlusion_query`.  [`EXT_occlusion_query_boolean`](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_occlusion_query_boolean.txt) only covered binary occlusion, so it can't be used for profiling.

But it does have [`EXT_disjoint_timer_query`](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_disjoint_timer_query.txt).  So apitrace could be enhanced to use this ES extension where available.

It's probably not a big task, but I'm afraid I don't have the time to pursuit nevertheless.

jrfonseca:
For completeness, another solution would be to run the trace with desktop GL, as ES traces should run well if the `GL_ARB_ES2_compatibility`/`GL_ARB_ES3_compatibility` extensions are supported.

lfrb:
I'm working on adding support using [EXT_disjoint_timer_query](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_disjoint_timer_query.txt). I have two related questions:

1) What should we do if a disjoint operation occurred (meaning the query resulting timestamp is not accurate and should not be used). Should we just not write anything or should we write `#INVALID`, `#DISJOINT` or whatever in the profiling log?

2) According to the spec, we should wait for the query result to be available before reading it, but I'm not seeing it in the code. Is it just a small mistake or is there a reason? (it's part of all three extensions: `EXT_timer_query`, `ARB_timer_query` and `EXT_disjoint_timer_query`)

```
        while (!available) {
            glGetQueryObjectiv(queries[1], GL_QUERY_RESULT_AVAILABLE, &available);
        }
[...]

PR #930, opened 2024-09-09
Add support for profiling with EXT_disjoint_timer_query

lfrb:
All GPU timestamp related metrics (and GPU profiling "-pgpu") are now available if the GPU supports EXT_disjoint_timer_query OpenGL extension.

Add "GPU Disjoint" metric for OpenGL backend (only possible to enable if "GPU Duration" is queried as well)

Fixes [689](https://github.com/apitrace/apitrace/issues/689)


jrfonseca:
LGTM.  Thanks.

PR #934, opened 2024-10-09
Fixes build errors on macOS

tanis2000:
This PR fixes an issue with CMake and retrace where it won't be built on macOS because on recent development tools the -sdk_version is no longer supported by `ld`.
This patch is taking the version from the system and feeding it to `ld` using -platform_version.

### DIFF
--- a/retrace/CMakeLists.txt
+++ b/retrace/CMakeLists.txt
@@ -142,9 +142,9 @@ if (WIN32 OR APPLE OR X11_FOUND)
                 #"-framework OpenGL" # CGL*
             )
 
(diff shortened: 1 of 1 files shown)


jrfonseca:
This change is not entirely correct -- it's the same as not passing any option at all --, but I've just tested on both Arm and X86, and this option indeed is not longer need.  That is we can just revert ff2e2d3f00397310560151a9f08e157d3bf40294

I'll push a revert of that commit shortly.  Thank you

tanis2000:
That sounds good

PR #936, opened 2024-11-22
Minor fixes and additions to ogl frametrim

Slaw6820:
During my work with apitrace I have encountered a small number of issues that were addressed with the following changes to the source code. Please review.


Slaw6820:
This is follow up to #905

PR #937, opened 2025-01-12
OVR_multiview extension support

amalon:
These changes add support for the OVR_multiview extension.
The first commit adds support for the gl[Named]FramebufferTextureMultiviewOVR() functions and new framebuffer attachment parameters.
The second uses the framebuffer attachment parameters to dump the specific subset of the layers of the base texture that are attached to an FBO (by getting the whole texture as for layered attachments, and then memcpy'ing the layer subsets from the pixel data).

Finally an unrelated commit fixes a recently introduced format warning due to the use of a size_t in a "%.*s" format.


retrace/glstate_images.cpp
```diff
@@ -792,6 +797,18 @@ dumpActiveTextureLevel(StateWriter &writer, Context &context,
         } else {
             glGetTexImage(target, level, format, type, image->pixels);
         }
+
+        if (num_layers > 0 && num_layers < desc.depth) {
```
jrfonseca, line 801: Shouldn't it be `num_layers <= desc.depth`?

jrfonseca (commented):
Otherwise looks good to me.

retrace/glstate_images.cpp
```diff
@@ -1727,10 +1744,20 @@ dumpFramebufferAttachment(StateWriter &writer, Context &context, GLenum target,
         glGetFramebufferAttachmentParameteriv(target, attachment,
                                               GL_FRAMEBUFFER_ATTACHMENT_LAYERED,
                                               &layered);
-        if (layered &&
-            isGeometryShaderBound(context)) {
+
+        // OVR_multiview
+        GLint first_view = 0;
```
jrfonseca, line 1749: Sorry.  Somehow my comment here got lost.

Have you tested this on a system that doesn't support `GL_OVR_multiview`?  I suspect this will trigger GL errors and cause troubles.

This code should be predicated by a `GL_OVR_multiview` extension check,

amalon: Yes, good spot. I had assumed the errors would be harmlessly ignored, but if I use glFramebufferTexture() and run without OVR_multiview, and then replay up to a draw command, GL_INVALID_ENUMs show up multiple times in the error log.
I'm guessing from a quick dig that OVR_multiview should be a new bit in struct Features, that can then be accessed from the context object.

amalon:
I've clarified the memcpy condition comments, and made the new glGetFramebufferAttachmentParameteriv calls conditional on a new OVR_multiview bit in struct Features.

jrfonseca:
Thanks.

PR #938, opened 2025-01-21
retrace: Use right API calls when using EXT_disjoint_timer_query

lfrb:
Even though the Mesa drivers are aliasing the EXT API calls with the unsuffixed versions, it doesn't work with all drivers (e.g. libmali). Selecting the right call to use depending on the supported extensions makes it work universally.

### DIFF
--- a/retrace/glretrace_main.cpp
+++ b/retrace/glretrace_main.cpp
@@ -205,6 +205,34 @@ getTimeFrequency(void) {
 }
 
 static inline void
+queryTimestamp(GLuint query) {
+    if (supportsDisjoint) {
+        glQueryCounterEXT(query, GL_TIMESTAMP_EXT);
+    } else {
+        glQueryCounter(query, GL_TIMESTAMP);
+    }
+}
+
+static inline int64_t
+getQueryResult(GLuint query) {
+    int64_t result = 0;
+
+    if (supportsDisjoint) {
+        glGetQueryObjecti64vEXT(query, GL_QUERY_RESULT_EXT, &result);
+    } else if (supportsTimestamp) {
+        glGetQueryObjecti64v(query, GL_QUERY_RESULT, &result);
+    } else if (supportsElapsed) {
+        glGetQueryObjecti64vEXT(query, GL_QUERY_RESULT, &result);
+    } else {
+        uint32_t result32;
+        glGetQueryObjectuiv(query, GL_QUERY_RESULT, &result32);
+        result = static_cast<int64_t>(result32);
+    }
+
+    return result;
+}
+
+static inline void
 getCurrentVsize(int64_t& vsize) {
     vsize = os::getVsize();
 }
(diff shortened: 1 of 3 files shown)


retrace/glretrace_main.cpp
```diff
@@ -204,6 +204,34 @@ getTimeFrequency(void) {
...
+static inline int64_t
+getQueryResult(GLuint query) {
+    int64_t result = 0;
+
+    if (supportsDisjoint) {
+        glGetQueryObjecti64vEXT(query, GL_QUERY_RESULT_EXT, &result);
+    } else if (supportsTimestamp) {
+        glGetQueryObjecti64v(query, GL_QUERY_RESULT, &result);
+    } else if (supportsElapsed) {
+        glGetQueryObjecti64vEXT(query, GL_QUERY_RESULT, &result);
```
kusma, line 225: This is the same case as the `supportsDisjoint`-case; GL_QUERY_RESULT_EXT and GL_QUERY_RESULT has the same value...

retrace/glretrace_main.cpp
```diff
@@ -204,6 +204,34 @@ getTimeFrequency(void) {
     }
 }
 
+static inline void
+queryTimestamp(GLuint query) {
+    if (supportsDisjoint) {
+        glQueryCounterEXT(query, GL_TIMESTAMP_EXT);
+    } else {
+        glQueryCounter(query, GL_TIMESTAMP);
+    }
```
kusma, line 213: Not sure if it matters, but `GL_TIMESTAMP_EXT` and `GL_TIMESTAMP` has the same value. The only thing that really matters here is `glQueryCounter` vs `glQueryCounterEXT`...

It also seems a bit backwards to prefer an extension function even if a core function is supported.... 

perhaps something like this?

```c++
static inline void
queryCounter(GLenum id, GLenum pname) {
   if (currentProfile.versionGreaterOrEqual(glfeatures::API_GL, 3, 3)) {
      glQueryCounter(id, target)
   } else
      assert(supportsDisjoint);
      glQueryCounterEXT(id, target);
   }
}
```

Dunno, perhaps I'm nit-picking a bit too much here.

retrace/glretrace_main.cpp
lfrb: Yeah, the issue is that:

```
  supportsTimestamp = suppotrsDisjoint | GL >= 3.3 | has_ext(ARB_timer_query)
  supportsElapsed = supportsTimestamp | has_ext(GL_EXT_timer_query)
```

so the other way of writing that would be

```
  if (supportsTimestamp && !supportsDisjoint)
    glGetQueryObjecti64v()
  else if (supportsElapsed)
    glGetQueryObjecti64vEXT
```

I thought the current version might be less confusing.

retrace/glretrace_main.cpp
lfrb: `glQueryCounter` is not only supported on OpenGL >= 3.3 though, it's also supported with `ARB_timer_query` on top of GL 3.2. So that's what `supportsTimestamp` is for...except it also includes `supportsDisjoint`. So we're back with the same issue than above.

I guess another solution would be to add new `bool` members to determine which API to use. Not sure which one is cleaner.

jrfonseca: I agree favoring the EXT over the core functions is odd. 

If we want to clean this up a bit better, then I think the ideal would be for us to:
- rename the `supportsFoo` variables with actual extension names, like we do in https://github.com/apitrace/apitrace/blob/d79087c8764608b544cab2c5fedb61859225e5fd/helpers/glfeatures.hpp#L161-L173 .  (In fact, we could consider moving these bools flags there.) 
- use the extension flags to determine which entry point to call

We could also consider using function pointers instead of if-then-elses.

But I'm ok merging as is and clean up subsequently.

PR #947, opened 2025-06-08
fix snappy detection, use build_testing for gtest detection too

PPN-SD:
Use config mode for snappy to prevent conflict with qtwebengine.
otherwise a conflict may occur with /usr/lib64/cmake/Qt6/findSnappy.cmake
See https://bugs.gentoo.org/951977

use BUILD_TESTING for gtest detection too

### DIFF
--- a/thirdparty/CMakeLists.txt
+++ b/thirdparty/CMakeLists.txt
@@ -59,10 +59,12 @@ if (CMAKE_EXECUTABLE_FORMAT STREQUAL "ELF")
     include_with_scope (libbacktrace.cmake)
 endif ()
 
-# We use non-standard C++ flags, so we can't just use GTest's CMakeLists.txt
-if (NOT GTEST_FOUND)
-    message (STATUS "Using bundled GTest")
-    include_with_scope (gtest.cmake)
+if (BUILD_TESTING)
+    # We use non-standard C++ flags, so we can't just use GTest's CMakeLists.txt
+    if (NOT GTEST_FOUND)
+        message (STATUS "Using bundled GTest")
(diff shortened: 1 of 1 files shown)


thirdparty/CMakeLists.txt
```diff
@@ -59,10 +59,12 @@ if (CMAKE_EXECUTABLE_FORMAT STREQUAL "ELF")
     include_with_scope (libbacktrace.cmake)
 endif ()
 
-# We use non-standard C++ flags, so we can't just use GTest's CMakeLists.txt
-if (NOT GTEST_FOUND)
-    message (STATUS "Using bundled GTest")
-    include_with_scope (gtest.cmake)
+if (BUILD_TESTING)
+  # We use non-standard C++ flags, so we can't just use GTest's CMakeLists.txt
```
jrfonseca, line 63: Indentation is off: please use 4 spaces.

jrfonseca (approved):
Just one cosmetic request.  Otherwise sounds and  looks good.  Thanks.

thirdparty/CMakeLists.txt
jrfonseca: Perhaps I wasn't clear: I meant 4 spaces _ per indentation level_, as per surrounding code.

You might also want to install an editorconfig plugin -- https://github.com/apitrace/apitrace/blob/master/.editorconfig#L12 -- for whatever editor you use, so these things are done automatically.

PPN-SD: I missed editorconfig, just installed with vim thanks !

PR #952, opened 2025-06-22
D3D8/9: Decode AL16 and R16 FOURCC formats

WinterSnowfall:
This PR aims to do two things, actually:
- Add support for the NULL format to d3d8, since it is the only "vendor hack" FOURCC from D3D9 that is being advertised (via `CheckDeviceFormat`) and accepted by D3D8 as well (I've checked this both on modern drivers and on age-accurate drivers/multiple AGP cards from the mid 2000s).

- Add partial support for the undocumented AL16 and R16 FOURCC formats, which a surprising amount of D3D8 and early D3D9 games use as part of `CheckDeviceFormat` calls. I am not entirely sure what is the story behind these formats, but they are common enough for a proper decoding to be warranted, as they show up in quite a few of the D3D8/9 traces I've captured. Neither are actually supported by any official drivers as far as I know, however WineD3D does allow them to be used in conjunction with `CreateImageSurface` (D3D8) / `CreateOffscreenPlainSurface` (D3D9).

I don't think I've missed anything, but that being said would more than appreciate a second pair of eyes, as I'm rather new to the codebase.

Edit: On second thought, I've made the assumption that the AL16 FOURCC hints at an A16L16 32-bit format (since A8L8 already exists).


jrfonseca (changes requested):
Indeed it seems old apps check for them, but nobody knows any drivers who implement them.

Per https://github.com/wine-mirror/wine/blob/e9dcdf38041a1911d37117e0219dced5ba5dbf7f/dlls/wined3d/utils.c#L149 WINE does treat AL16 as 16+16=32 bpp format.

WinterSnowfall:
> WINE does treat AL16 as 16+16=32 bpp format.

Yeah, makes sense in retrospect. Should be fine now.

jrfonseca:
Thanks.

PR #956, opened 2025-07-14
trace: reset locked array count when unlock.

llyzs:
This fixes a memory corruption that happens at `_trace_user_arrays` during subsequent draw calls after glUnlockArraysEXT called but the lockedArrayCount is not reset.


jrfonseca:
Good catch.

I'm surprised `GL_EXT_compiled_vertex_array` is still being used though as opposed to VBOs.  I suppose it's an old app.

PR #959, opened 2025-08-23
d3d9retrace: Also adjust the presentation interval on device reset

WinterSnowfall:
Fixing up the PresentationInterval / FullScreen_PresentationInterval also needs to be considered during device reset, analogous to how it's being handled on device creation.

P.S.: The problem was observed in the trace attached to this issue, for reference: https://github.com/doitsujin/dxvk/issues/5166


jrfonseca (approved):
Makes sense.  Thanks.

PR #961, opened 2025-09-23
Re-enable and fix GL window system cleanup

anholt:
We have a number of apitraces that we'd like to use for driver performance work with perfetto or MESA_GPU_TRACES=print_*, but we can't because a lack of cleanup of the window system means that the output gets arbitrarily truncated.

I've now confirmed that my screen gets destroyed on an EGL and a GLX trace from traces-db, with and without waffle.  TBD: d3d.


jrfonseca:
Makes sense.  Thanks

PR #965, opened 2025-10-27
Add an option to convert a trace to C code

KonstantinSeurer:
This is a tool that is similar to the `gfxrecon-tocpp` tool with the main goal to speed up replay for performance and correctness testing.

```
Adds the --generate-c option that specifies a directory where glretrace
will generate a C/C++ meson project that contains the API calls of the
trace. Most of thge codegen is implemented in retrace_codegen.cpp and
OpenGL specific quirks are located in glretrace_codegen.cpp.

For replaying the trace, one simply points glretrace to the .so that can
be built from the output. glretrace will try to load trace::Blob data
from the .so.data file in the same directory as the .so.

API calls that are non-trivial to replay like glx will still use the
normal trace::Call codepath.

The .so replay only supports a subset of the options for nor al replay.
Most importantly snapshots work, but only at frame granularity.
```


zmike:
As the overseer of this project I guess I'll reply here:

First, there were a few goals in mind with adding this feature:
* Mesa (and other projects/companies) uses apitrace in CI for correctness testing in many, many jobs across many drivers. These CI jobs all have a real cost in dollars to run, and they take a long time. Running compiled traces is significantly faster, which means CI jobs become cheaper in real world dollars. It also means CI becomes less of a blocker and merges can trigger quicker, not to mention the reliability improvements which come from having those shorter jobs
* Mesa wants to use apitrace for performance testing across drivers, but this is not currently viable due to: 1) inconsistencies with playing single-frame traces 2) the lengthy runtime of multi-frame traces. Compiling traces brings consistency here and enables more use of apitrace with existing traces
* Driver developers need a better solution for doing holistic performance testing; while it's possible to have apitrace print GPU frametimes with retrace, making changes and then running a game/app to gauge CPU performance over and over is frustrating. With compiled traces, developers can use existing tools in a new way to improve their workflows

Now as to your above points, I've read all your discussions regarding this type of functionality, but I don't believe you've ever given a convincing refutation of it. In essence, your arguments boil down to:
* if is the goal, use LLVM IR; otherwise use python
* if editability is the goal, add functionality to enable creating a trace from text (and also improve trimming)
* this adds a maintenance burden

Neither of the first two statements reasonably addresses the goals that people have in generating C/C++ out of traces:
* Speed is obviously a factor;  as noted in [Konstantin's blog post](https://www.supergoodcode.com/apitrace-goes-vroom/), retrace is very slow. This isn't something that can be truly solved at the parser/retrace level since the act of parsing a trace introduces CPU overhead, which means a different sort of solution is required, ideally one which doesn't break compatibility with existing traces by e.g., changing the trace format
  * LLVM IR is definitely not a solution. As a llvmpipe developer I **know** you're well aware of the titanic struggle that is writing effective LLVM code and keeping up with all the version idiosyncracies/bugs. Nobody wants to add more LLVM into their workflows, and the Venn diagram of `(people who can/will potentially contribute to apitrace)` and `(people who know/enjoy LLVM IR)` is vanishingly tiny. If the burden of maintenance is a real concern, I'm surprised this is something you would even consider, as it would add a massive new dependency and would almost certainly guarantee the functionality is immediately orphaned
  * Python is not as fast as C/C++ without being compiled down, and at that point why not just use C/C++ anyway? It isn't as though apitrace doesn't already have generated C/C++ code
  * Originally the generated code was pure C++, but compiling C++ is much slower than C, to the point of being unusable for this purpose. A compromise was made for usability, and there was no technical benefit to having all the GL function calls in C++ files anyway
  * The precedent for C/C++ trace code was already set years ago by [ANGLE](https://chromium.googlesource.com/angle/angle/+/main/doc/CaptureAndReplay.md), which does exactly this; anyone coming from the mobile space expects this functionality
* Editability is also nice, and in the graphics driver space, C is the language people are most familiar with; it isn't a coincidence that "convert traces to C" is a request which keeps popping up instead of "convert traces to Python/LLVM"
  * Conflating this work with improvements to trimming is not a sound argument; they are separate functionalities with separate goals, and one does not impede the other. We've already been independently discussing work that may be done here since this is another problem area for us and many others

With all that said, I can sympathize with adding additional burdens when you're the sole maintainer of a project. I don't have a magic wand to wave that gets you a fleet of engineers hacking away at all the bugs, and nobody is promising to stick around and work on stuff regularly, but removing features and rejecting PRs adding much-requested functionality is not a path towards growing the ecosystem either. Apitrace is a great project, one which is overlooked because it mostly "just works". Maybe if people see that it can still become even better, they'll grow more interested in contributing. Maybe not. Either way, I think the approach here remains consistent with the rest of the codebase, and it succeeds without introducing new dependencies or making a mess of things. That's probably the best you can hope for.


My expectation is that people will still create trace files as per usual. Compiling traces is an investment in both time and disk space, so retrace will still be the primary means by which most people interact with traces. Having the option to compile traces, however, opens up new use cases for apitrace in both the corporate and hobbyist spheres, and it brings us to parity with ANGLE in this regard, which is sure to make the project more interesting to many consumers.

(oops double posted)

jrfonseca:
Sorry for the silence.  I've been quite busy and I also wanted the time to mull over this.

Thanks for the explanations.  I think I have a better handle where you're coming from.

> This isn't something that can be truly solved at the parser/retrace level since the act of parsing a trace introduces CPU overhead, which means a different sort of solution is required, ideally one which doesn't break compatibility with existing traces by e.g., changing the trace format

I think changing the trace format deserves more consideration.

The current trace format was designed to be compact, and forces reparsing to happen in a serial fashion.  Parsing is slow because it's effectively akin to drinking an ocean through a straw.

If the goal is have very fast parse/replay, and disk space is less of an issue, then a new trace format more amenable to mmap'ing (or at least parallel decompression and parsing) could go a long way to alleviate these things.

We can then provide a CLI command (like `apitrace repack`) which converts older format to the new.  It would be an one-off cost and no need to retrace applications.

> LLVM IR is definitely not a solution. As a llvmpipe developer I **know** you're well aware of the titanic struggle that is writing effective LLVM code and keeping up with all the version idiosyncracies/bugs. Nobody wants to add more LLVM into their workflows, and the Venn diagram of _(people who can/will potentially contribute to apitrace)_ and _(people who know/enjoy LLVM IR)_ is vanishingly tiny. If the burden of maintenance is a real concern, I'm surprised this is something you would even consider, as it would add a massive new dependency and would almost certainly guarantee the functionality is immediately orphaned

That's actually very far from my perspective.  The use of LLVM IR in llvmpipe never gave me much struggle beyond having to hunt down memory leaks.  Nothing like the grief I had dealing with half-baked forks/rewrites of apitrace: explaining repeatedly why something is a bad idea, taking flak when refusing to merge it PR when it looks so promising for this or that particular example, dealing with the broken pieces when accepting the merge...

This Venn diagram you imagine is theorical.  In practice it has always been only myself, and I could imagine myself maintaining LLVM IR because it provides a path for complete _replacement_ of C code generation via Python scripts (one way to rule them all.)  In fact, if I were to start Apitrace from today, using LLVM IR is how I'd built retrace functionality from the outset.  Whereas I don't see myself maintaining a retrace code path in _addition_ to the current one.

But this is is moot either way: Apitrace exists as-is today, and because I no longer use apitrace for my day-to-day work, I won't be rewriting it or even changing it much.

> Originally the generated code was pure C++, but compiling C++ is much slower than C, to the point of being unusable for this purpose. A compromise was made for usability, and there was no technical benefit to having all the GL function calls in C++ files anyway

I see.  But D3D is C++, and the same motives behind doing this for GL apply to D3D too.  So not only I'd have this extra code path, but it would be GL specific too.

> [...] removing features and rejecting PRs adding much-requested functionality is not a path towards growing the ecosystem either

True.  But that has never seemed to have made a difference: I have accepted such PRs in the past and then regretted, as people didn't stick around regardless.


So in the end, this all boils down to the following mental calculation: does the net benefit for others of accepting offset the net cost for me.

But I'm afraid the answer in this case is no.  I might be wrong, but this still feels like scratching a very special itch to me.  Sorry.


And to avoid further disappointments, I'll try to make it clear the dividing lines in my mind: imagine apitrace is a tricycle which rests on three wheels: tracing, retracing, and state inspection.  I'm all for work which speeds up any of these wheels, or replaces one of the wheels with one better.  I cannot accept work which effectively adds extra wheels: stateful trimming, exporting to C, start tracing mid application execution, etc.  I've been there, done that, and invariably regretted it.

baskanov:
> I see. But D3D is C++, and the same motives behind doing this for GL apply to D3D too. So not only I'd have this extra code path, but it would be GL specific too.

D3D API is based on COM, which can be used both from C++ and C.

jrfonseca:
> D3D API is based on COM, which can be used both from C++ and C.

True.  I know from experience that D3D C bindings are often broken/incomplete from lack of use/testing, and using COM from C is cumbersome due to lack of compiler assisted casting/ref-counting, but it's indeed possible in theory one could spit out C that uses D3D.

fooishbar:
@KonstantinSeurer I've pushed a bunch of fixes to `fooishbar/apitrace:codegen-fixes`.

KonstantinSeurer:
> @KonstantinSeurer I've pushed a bunch of fixes to fooishbar/apitrace:codegen-fixes.

Thanks, I pulled your patches into my branch.

PR #973, opened 2025-12-23
glretrace: fix glReadPixels crash when width/height is large

yuq:
When replay glReadPixels with width=8192 and height=4096, original code just hard code pixel size to 64 bytes which need to allocate 2G buffer and throw std::length_error.

In fact we can get accurate required size to reduce the buffer to 128M when pixel size is 4 bytes in most cases.

### DIFF
--- a/retrace/glretrace.py
+++ b/retrace/glretrace.py
@@ -227,7 +227,7 @@ class GlRetracer(Retracer):
         print(r'    if (currentContext && currentContext->features().pixel_buffer_object) {')
         print(r'        glGetIntegerv(GL_PIXEL_PACK_BUFFER_BINDING, &_pack_buffer);')
         print(r'    }')
-        print(r'     std::vector<char> buffer;')
+        print(r'    std::vector<char> buffer;')
         print(r'    if (!_pack_buffer) {')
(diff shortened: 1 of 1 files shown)


retrace/glretrace.py
```diff
@@ -241,7 +241,10 @@ def overrideArgs(self, function):
         elif function.name == "glReadPixels":
             print(r'     GLsizei _w = call.arg(2).toSInt();')
             print(r'     GLsizei _h = call.arg(3).toSInt();')
-            print(r'     buffer.resize(_w * _h * 64);')
+            print(r'     GLenum _format = call.arg(4).toSInt();')
+            print(r'     GLenum _type = call.arg(5).toSInt();')
+            print(r'     unsigned _pixel_size = _gl_format_size(_format, _type) / 8;')
```
jrfonseca: It should be actually something like below to deal accurately with single-bit formats.
```python
            print(r'     unsigned _bits_per_pixel = _gl_format_size(_format, _type);')
            print(r'     unsigned _stride = (_w * _bits_per_pixel + 7) / 8;')
            print(r'     buffer.resize(_h * _stride);')
```

But I'll fix it in a subsequent change.

PR #982, opened 2026-09-17
Support replaying d3d9 traces via --driver=d3d9on12

hushz-brcm:
This change introduces a new `--driver=d3d9on12` option to d3dretrace, allowing pre-recorded D3D9 traces to be replayed through the D3D9on12 mapping layer. When using this driver, D3D9 API calls are translated into D3D12 calls under the hood via `Direct3DCreate9On12`. This feature provides a lightweight way to develop and validate Direct3D 12 User Mode Drivers (UMDs) by leveraging existing D3D9 traces.


jrfonseca:
@hushz-brcm ,

MinGW headers are indeed often incomplete and we need to compensate in apitrace source code.  Still, we favour using MSVC system headers where they are available.  So please wrap the https://github.com/apitrace/apitrace/pull/982/changes/28a7fdda94ece675dcad958bd732da1408f53cde change in `#ifdef/#ifndef  __MINGW32__ ... #endif`, like:

```c++

#ifndef __MINGW32__
#include <d3d9on12.h>
#endif

...

#ifdef __MINGW32__
typedef struct D3D9ON12_ARGS {
...
#endif
```

Otherwise looks ready to go.

hushz-brcm:
Hi Jose, thanks for the feedback. I have addressed them and rebased to top of tree.

jrfonseca:
Merged.  Thanks, Husheng.

PR #983, opened 2026-09-21
glretrace: Honor GL_PACK_* when sizing glReadPixels replay buffers

luokn:
When no `GL_PIXEL_PACK_BUFFER` is bound, retrace allocates a temporary CPU
buffer and redirects `glReadPixels` at that pointer.  That buffer was sized as
a compact `width * height` rectangle and ignored pixel pack state:

- `GL_PACK_ALIGNMENT`
- `GL_PACK_ROW_LENGTH`
- `GL_PACK_SKIP_PIXELS`
- `GL_PACK_SKIP_ROWS`

Drivers write according to those parameters, so a correct replay could overflow
the retrace buffer.

A real example from Firefox / Babylon.js GLES traces:

```text
glPixelStorei(GL_PACK_ALIGNMENT, 4)
glPixelStorei(GL_PACK_ROW_LENGTH, 800)
glReadPixels(..., width=797, height=774, GL_RGBA, GL_UNSIGNED_BYTE, ...)
```

For `RGBA8` that is `774 * 800 * 4 = 2,476,800` bytes written, but retrace only
allocated `774 * 797 * 4 = 2,467,512` bytes.

## Changes

- Size the replay buffer from the current pack layout.
- Query `GL_PACK_ROW_LENGTH` / `GL_PACK_SKIP_*` only where they exist:
  desktop GL, GLES 3.0+, or `GL_NV_pack_subimage`.
- GLES 1/2 without that extension still query `GL_PACK_ALIGNMENT` only, so they
  do not generate `GL_INVALID_ENUM` or pollute later `glGetError()` results.
- Add `_glReadPixels_size`, analogous to `_glTexImage2D_size`, with
  `can_pack_subimage()` generated by `glretrace.py`.

This is a destination-buffer size, not `_gl_image_size()`.  Unpack sizing is
for reading a source image and can omit padding on the last row.  Pack sizing
must cover a full row stride, because drivers may write that padding.

## Testing

Built `glretrace`, `eglretrace`, `glxtrace`, and `egltrace`.  Ran the existing
`check` target.


jrfonseca:
Looks great.  Thanks

